PR #151, opened 2024-02-09
Add option to disable normalize embeddings after calculating emphasis

hako-mikan:
## Description

In SDXL, there have been reports of artifacts occurring when combining prompts or using long prompts. I have discovered the cause of this phenomenon. It is the normalization that occurs after the calculation of emphasis. During this normalization, artifacts arise due to division by values close to zero and the inversion of signs. This normalization process dates back to the SD1.X era, where it seems to work well. However, in certain models of XL, it appears to have the opposite effect. Therefore, I will add an option to toggle this normalization on and off.
The detailed explanation can be found [here](https://civitai.com/articles/4044) or [here(Japanese)](https://note.com/hakomikan/n/nb6dd68a1bd9e)

## Checklist:

- [x] My code passes [tests](https://github.com/AUTOMATIC1111/stable-diffusion-webui/wiki/Tests)


lllyasviel:
yes prs related to webui should go upstream

Issue #110, opened 2024-02-07
[Bug]: Built in control net extention throws exception when used via API


mikedreambooth:
### Checklist

- [x] The issue exists after disabling all extensions
- [x] The issue exists on a clean installation of webui
- [ ] The issue is caused by an extension, but I believe it is caused by a bug in the webui
- [X] The issue exists in the current version of the webui
- [x] The issue has not been reported before recently
- [ ] The issue has been reported before but has not been fixed yet

### What happened?

When making an api request with Controlnet modules defined in the payload controlnet will throw an exception. (Same payload works with standard webgui + controlnet extention )

`File "D:\stable-diffusion-webui-forge\extensions-builtin\sd_forge_controlnet\scripts\controlnet.py", line 97, in get_enabled_units
        enabled_units = [x for x in units if x.enabled]` 

### Steps to reproduce the problem

Make a post request with a payload which includes control net modules

### What should have happened?

Controlnet args should be parsed correctly from the request.

### What browsers do you use to access the UI ?

Other

### Sysinfo

[sysinfo-2024-02-07-17-20.json](https://github.com/lllyasviel/stable-diffusion-webui-forge/files/14197195/sysinfo-2024-02-07-17-20.json)

### Console logs

```Shell
VAE dtype: torch.bfloat16
Using pytorch cross attention
ControlNet preprocessor location: D:\stable-diffusion-webui-forge\models\ControlNetPreprocessor
Loading weights [15012c538f] from D:\stable-diffusion-webui-forge\models\Stable-diffusion\realisticVisionV51_v51VAE.safetensors
2024-02-07 14:27:33,043 - ControlNet - INFO - ControlNet UI callback registered.
Running on local URL:  http://127.0.0.1:7860

To create a public link, set `share=True` in `launch()`.
model_type EPS
UNet ADM Dimension 0
Startup time: 13.0s (prepare environment: 3.1s, import torch: 4.0s, import gradio: 1.2s, setup paths: 1.1s, initialize shared: 0.1s, other imports: 0.8s, load scripts: 1.3s, create ui: 0.6s, gradio launch: 0.4s, add APIs: 0.4s).
Using pytorch attention in VAE
Working with z of shape (1, 4, 32, 32) = 4096 dimensions.
Using pytorch attention in VAE
extra {'cond_stage_model.clip_l.text_projection', 'cond_stage_model.clip_l.logit_scale'}
loaded straight to GPU
To load target model BaseModel
[...]

PR #162, opened 2024-02-10
ControlNet API

huchenlei:
## Description

Closes https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/110.
This PR adds back API support for ControlNet.

List of things achieved in this PR
- Unittests for ControlNet API
- Add `processed.extra_images` to API response
- Fix the default value for many `ControlNetUnit` fields. The extra tailing `,` is making default value tuple of single item
- Add back `save_detected_map` API only option
- Parse ControlNet API payload

## Checklist:


lllyasviel (approved):

huchenlei:
Seems like inference with real SD model on CPU is too slow for ControlNet tests. Let me find a way to load empty model for tests.

huchenlei:
Temporarily disable ControlNet tests in CQ. All tests passed locally.

mikedreambooth:
@huchenlei in regards to #110 I tested an api request and the base64 image is not converted to an nd array correctly so controlnet throws still.

the function added to convert from dict to UiControlNetUnit in external_code.py does not reference the image correctly unit.image instead of unit.image['image'] so the type check fails and it remains base64

PR #165, opened 2024-02-10
Accept image paths via ControlNet API

sunadarake:
## Description

when using ControlNet via API, it is necessary to convert images to base64. 

This PR modifies it to accept image paths when sending images via the API.

## Checklist:

### DIFF
--- a/extensions-builtin/sd_forge_controlnet/lib_controlnet/external_code.py
+++ b/extensions-builtin/sd_forge_controlnet/lib_controlnet/external_code.py
@@ -5,6 +5,7 @@ import numpy as np
 from modules import shared
 from lib_controlnet.logging import logger
 from lib_controlnet.enums import InputMode, HiResFixOption
+from lib_controlnet.utils import image_dict_from_any
 from modules.api import api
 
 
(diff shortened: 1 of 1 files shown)


huchenlei:
I think the original API design makes a lot of confusion on how should we specify input image / mask. The number of input semantics increased the code complexity / maintainance cost. In forge, we would like to keep it simple, i.e. only support one most used semantic. `image_dict_from_any` is planned to be removed.

@light-and-ray also made a valid point about the security concern, as if you run forge as a remote server. Accepting file name would expose all files on the server to the user.

It is not hard to add following function to your script to read the local image and convert it to base64 string.
```python
def read_image(img_path: Path) -> str:
    img = cv2.imread(str(img_path))
    _, bytes = cv2.imencode(".png", img)
    encoded_image = base64.b64encode(bytes).decode("utf-8")
    return encoded_image
```

huchenlei:
Documentation on web api usage added: https://github.com/lllyasviel/stable-diffusion-webui-forge/wiki/ControlNet-Web-API

sunadarake:
@light-and-ray @huchenlei 

Thank you for reviewing.
I hadn't considered security aspects as I was only assuming the use of the WebUI in a local environment.
I will withdraw this PR.

PR #309, opened 2024-02-18
Only feed image embed to instant id ControlNet unit

huchenlei:
## Description

Current implementation of `instant_id_modifier` replaces controlnet's crossattn input unconditionally regardless if the unit is an instsant id unit or not.

This PR adds a condition on `instant_id_modifier`, which only applys the overwrite of crossattn input when the unit is a instant id.

There is still an issue when there need to be multiple faces controlled by instant id on canvas. Current mechanism will make the face set later overwrite the face set eariler.

We need to find a way to pair ipadapter unit and controlnet unit. The solution in sd-webui-controlnet is to make a convention that instant id controlnet's previous unit is always the pairing instant id ipadapter unit.

Note: I do not find a good control type to test this. Canny does not seem to respond differently when giving image embed as crossattn vs giving text embed as crossattn.

## Checklist:

### DIFF
--- a/extensions-builtin/sd_forge_controlnet/tests/web_api/generation_test.py
+++ b/extensions-builtin/sd_forge_controlnet/tests/web_api/generation_test.py
@@ -169,3 +169,22 @@ def test_lama_outpaint():
             "resize_mode": "Resize and Fill",  # OUTER_FIT
         },
     ).exec()
+
+
+@disable_in_cq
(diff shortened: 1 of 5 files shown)


BurnZeZ:
To be clear, is this a bugfix, or an enhancement?

huchenlei:
This is a bugfix.

Issue #51, opened 2024-02-06
[Bug]:  Live Preview image not shown at full quality despite setting to "Full" (Config:   "show_progress_type": "Full")


CCpt5:
### Checklist

- [X] The issue exists after disabling all extensions
- [X] The issue exists on a clean installation of webui
- [ ] The issue is caused by an extension, but I believe it is caused by a bug in the webui
- [X] The issue exists in the current version of the webui
- [X] The issue has not been reported before recently
- [ ] The issue has been reported before but has not been fixed yet

### What happened?

Preview image during generation is never "full quality" even if set as full. Compared all settings between my current A1111 setup (latest dev build) and Forge and they all match. Config.json also has "show_progress_type": "Full" for both, but A1111 does what it should and generates/displays the full image during generating - Forge displays a degraded image.

![65141](https://github.com/lllyasviel/stable-diffusion-webui-forge/assets/37972078/c10d46e5-80a3-44bb-bebd-a511e5411693)
![332131](https://github.com/lllyasviel/stable-diffusion-webui-forge/assets/37972078/6a697179-71ff-41fd-97d8-ff116b8aa589)

### Steps to reproduce the problem

set UI for full quality previews and check. 

### What should have happened?

full quality preview

### Sysinfo

Edit/Including:

<details><summary>Details</summary>
<p>

```
{
    "Platform": "Windows-10-10.0.19045-SP0",
    "Python": "3.10.6",
    "Version": "f0.0.9-latest-51-g5bea443d",
    "Commit": "5bea443d94f3a85f819cb8541c1bba0aac208d83",
    "Script path": "D:\\stable-diffusion-webui-forge",
    "Data path": "D:\\stable-diffusion-webui-forge",
    "Extensions dir": "D:\\stable-diffusion-webui-forge\\extensions",
    "Checksum": "e02ad33627cf67731f3fb6de7af4af9c257551750c868af4f8f63db81aaa30ef",
    "Commandline": [
        "launch.py",
        "--ckpt-dir",
        "e:\\stable Diffusion Checkpoints"
    ],
    "Torch env info": {
        "torch_version": "2.1.2+cu121",
        "is_debug_build": "False",
        "cuda_compiled_version": "12.1",
        "gcc_version": null,
        "clang_version": null,
        "cmake_version": null,
        "os": "Microsoft Windows 10 Pro",
        "libc_version": "N/A",
[...]

PR #408, opened 2024-02-26
Allow previews at full quality

catboxanon:
## Description

Closes https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/51

To me this seems like a leftover that was accidentally left in after gutting the lowvram code from upstream. https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/18819723c1d727a5a96224011607963d44a04d36/modules/sd_samplers_common.py#L44

## Checklist:

- [x] My code passes [tests](https://github.com/AUTOMATIC1111/stable-diffusion-webui/wiki/Tests)

### DIFF
--- a/modules/sd_samplers_common.py
+++ b/modules/sd_samplers_common.py
@@ -41,8 +41,6 @@ def samples_to_images_tensor(sample, approximation=None, model=None):
 
     if approximation is None or (shared.state.interrupted and opts.live_preview_fast_interrupt):
         approximation = approximation_indexes.get(opts.show_progress_type, 0)
-        if approximation == 0:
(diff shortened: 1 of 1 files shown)


lllyasviel:
This is intentional because full preview significantly drop performance by causing GPU memory fragments even for full vram mode with both unet and vae in GPU.

if you put some thread debug tools on upstream 'samples_to_images_tensor' function you will find that thread number changes a lot each generation or sometimes even each iterations. This is causing some thread GPU context management overheads on pytorch

I am still investigating the solution but forge will not add it back before it is safe to do that

PR #475, opened 2024-03-04
Fix alphas cumprod

continue-revolution:
Without this PR, I cannot reproduce any result from auto upstream.
 
The reason is that ComfyUI init alphas cumprod at model loading time instead of sample time (auto does this) so we need another step.

Another change is that ComfyUI always does "[quantize](https://github.com/crowsonkb/k-diffusion/blob/master/k_diffusion/external.py#L70)" when it convert sigma to t, but A1111 never does "quantize" from my observation.

Comparison:
| A1111 v1.8.0 | Forge before this change | Forge after this change |
| --- | --- | --- |
|  |  |  |

### DIFF
--- a/ldm_patched/modules/model_sampling.py
+++ b/ldm_patched/modules/model_sampling.py
@@ -76,7 +76,13 @@ class ModelSamplingDiscrete(torch.nn.Module):
     def timestep(self, sigma):
         log_sigma = sigma.log()
         dists = log_sigma.to(self.log_sigmas.device) - self.log_sigmas[:, None]
-        return dists.abs().argmin(dim=0).view(sigma.shape).to(sigma.device)
+        low_idx = dists.ge(0).cumsum(dim=0).argmax(dim=0).clamp(max=self.log_sigmas.shape[0] - 2)
+        high_idx = low_idx + 1
+        low, high = self.log_sigmas[low_idx], self.log_sigmas[high_idx]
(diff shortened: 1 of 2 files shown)


continue-revolution:
close https://github.com/continue-revolution/sd-forge-animatediff/issues/9

strawberrymelonpanda:
Certainly a lot closer and would be a welcomed fix, I haven't had any luck with AnimateDiff in Forge compared to A1111.

It's interesting there's still some minute differences between the two examples - The most noticeable is in Forge her left hand (viewer's right) disappears earlier.

I believe there's some very small differences in the skirt wrinkles and perhaps the way the shadows 'fill in' on the top as the arm lowers as well.

lllyasviel (approved):

continue-revolution:
@strawberrymelonpanda this is because forge’s clip text encoder (which encodes your prompt into a tensor) also gives slightly different result compared to A1111. I am not sure about the reason. Most likely another comfy issue. I also observe slight difference when I do single image generation.

MUTED64:
Error "Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu" occurred after this PR merged. [Here is the traceback.](https://pastebin.ubuntu.com/p/bqKNtxZ9BD/)

lllyasviel:
this has some conflicts to memory management and i2i and will be reviewed by me later

PR #476, opened 2024-03-04
fix sampler device

continue-revolution:

PR #478, opened 2024-03-04
fix alphas cumprod

continue-revolution:
#475 but without modifying sigma to t method

this should only affect AnimateDiff

@lllyasviel can you approve this first?


lllyasviel:
let me take a look

continue-revolution:
this pr is now only related to alphas cumprod patch.

#476 is now only related to the "quantization".

lllyasviel (approved):

PR #487, opened 2024-03-04
Add New Controlnet Nromal preprocesser Dsine

sdbds:
## Description

Add New Controlnet Nromal preprocesser Dsine
extensions-builtin/forge_preprocessor_normaldsine
add get_pad in modules_forge/forge_util.py


huchenlei (commented):
Is it possible that we host normaldsine as a python package dependency just like depth_anything?

sdbds:
> Is it possible that we host normaldsine as a python package dependency just like depth_anything?

I think it's a good idea, it's easier to install using python package.

Issue #439, opened 2024-02-28
[Bug]: SD difference (script) sd-webui-IS-NET-pro cannot input images to controlnet


k52252467:
### Checklist

- [ ] The issue exists after disabling all extensions
- [ ] The issue exists on a clean installation of webui
- [X] The issue is caused by an extension, but I believe it is caused by a bug in the webui
- [X] The issue exists in the current version of the webui
- [X] The issue has not been reported before recently
- [ ] The issue has been reported before but has not been fixed yet

### What happened?

I'm in stable-diffusion-webui,
You can use the sd-webui-IS-NET-pro plug-in (script) normally,
I use his Txt_to_Image_multi_frame_rendering.py function,
Used in T2I,
But it cannot be used in stable-diffusion-webui-forge.
He can't seem to use Use another image as ControlNet input into ControlNet

This is the URL of the plugin
https://github.com/ClockZinc/sd-webui-IS-NET-pro

I know this may not be related to your script, but I also can't use the original ControlNet plugin in your stable-diffusion-webui-forge, it gets disabled after installation,

Does your built-in ControlNet not support certain scripts or plug-ins?

You can see in the LOG, it shows that I have used any image, but I use sd-webui-IS-NET-pro and it can Use another image as ControlNet input

![螢幕擷取畫面 2024-02-28 223052](https://github.com/lllyasviel/stable-diffusion-webui-forge/assets/130882375/6b5c97bc-4301-449f-9a3b-d8c8e87838b0)

In the picture, although I checked Use another image as ControlNet input,
But it can still be used normally in the original stable-diffusion-webui. Of course, I also tried to check it, but it still cannot be used in stable-diffusion-webui-forge.

simply put,
stable-diffusion-webui is used normally,
Not available in stable-diffusion-webui-forge

### Steps to reproduce the problem

Use sd-webui-IS-NET-pro plug-in (script) for T2I,
  Txt_to_Image_multi_frame_rendering.py,

He cannot input image into ControlNet

### What should have happened?

He should be able to use Use another image as ControlNet input to controlnet,
Because I need hundreds to thousands of pictures to convert at a time,
He should convert accordingly according to the picture number.

### Sysinfo

[sysinfo-2024-02-28-14-18.json](https://github.com/lllyasviel/stable-diffusion-webui-forge/files/14435254/sysinfo-2024-02-28-14-18.json)

### Console logs

```Shell
[...]

Issue #259, opened 2024-02-14
[Bug]: Control Net not working and throwing errors


shingenko:
### Checklist

- [ ] The issue exists after disabling all extensions
- [X] The issue exists on a clean installation of webui
- [ ] The issue is caused by an extension, but I believe it is caused by a bug in the webui
- [X] The issue exists in the current version of the webui
- [X] The issue has not been reported before recently
- [ ] The issue has been reported before but has not been fixed yet

### What happened?

I've only tried this using control net with CANNY and OPENPOSE, but in both cases I am getting the following error. It still generates an output but it looks like it's completely ignoring control net because of an error. The generated image has nothing to do with the control net input. The error is repeated for each attempted image generation.

```
*** Error running postprocess_batch_list: F:\StableDiffusion\stable-diffusion-webui-forge\extensions-builtin\sd_forge_controlnet\scripts\controlnet.py
    Traceback (most recent call last):
      File "F:\StableDiffusion\stable-diffusion-webui-forge\modules\scripts.py", line 854, in postprocess_batch_list
        script.postprocess_batch_list(p, pp, *script_args, **kwargs)
      File "F:\StableDiffusion\stable-diffusion-webui-forge\venv\lib\site-packages\torch\utils\_contextlib.py", line 115, in decorate_context
        return func(*args, **kwargs)
      File "F:\StableDiffusion\stable-diffusion-webui-forge\extensions-builtin\sd_forge_controlnet\scripts\controlnet.py", line 559, in postprocess_batch_list
        self.process_unit_after_every_sampling(p, unit, self.current_params[i], pp, *args, **kwargs)
    KeyError: 0

---
*** Error running process_before_every_sampling: F:\StableDiffusion\stable-diffusion-webui-forge\extensions-builtin\sd_forge_controlnet\scripts\controlnet.py
    Traceback (most recent call last):
      File "F:\StableDiffusion\stable-diffusion-webui-forge\modules\scripts.py", line 830, in process_before_every_sampling
        script.process_before_every_sampling(p, *script_args, **kwargs)
      File "F:\StableDiffusion\stable-diffusion-webui-forge\venv\lib\site-packages\torch\utils\_contextlib.py", line 115, in decorate_context
        return func(*args, **kwargs)
      File "F:\StableDiffusion\stable-diffusion-webui-forge\extensions-builtin\sd_forge_controlnet\scripts\controlnet.py", line 553, in process_before_every_sampling
[...]

PR #530, opened 2024-03-10
Fix for controlnet(integrated)'s api

papuSpartan:
## Description
Currently, sending an api request to controlnet integrated may result in issues. This is due to the base64 encoded images being accidentally left in base64 format. A check fails [here](https://github.com/papuSpartan/stable-diffusion-webui-forge/blob/01c24e44a624a1b85e125e5e35b3b572a3c78755/extensions-builtin/sd_forge_controlnet/lib_controlnet/external_code.py#L258) because `unit.image` (and later `unit.mask_image` also) are always of type `dict`. The original comparison looks to be checking if contents are type `str` to infer if they are encoded. In practice, this fails without checking the `image` key for each which should actually contain the b64 image string.

May resolve: #439, #259, papuSpartan/stable-diffusion-webui-distributed#25

## Checklist:

### DIFF
--- a/extensions-builtin/sd_forge_controlnet/lib_controlnet/external_code.py
+++ b/extensions-builtin/sd_forge_controlnet/lib_controlnet/external_code.py
@@ -261,8 +261,8 @@ class ControlNetUnit:
                 "image": img,
                 "mask": np.zeros_like(img),
             }
-        if isinstance(unit.mask_image, str):
-            mask = np.array(api.decode_base64_to_image(unit.mask_image)).astype('uint8')
+        if isinstance(unit.mask_image['image'], str):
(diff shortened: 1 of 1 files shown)


papuSpartan:
Had a misunderstanding, unit formatting for the api should be like this per #162 
```
{
  "controlnet": {
    "args": [
      {
        "image": "<base64string>",
        "image_mask": "<base64string>",
        ...
      },
      {
        "image": "<base64string>",
        "image_mask": "<base64string>",
        ...
      }
    ]
  }
}
```

`unit.image: str` is analogous to `unit.input_image: str` in mikubill's controlnet but I'm unsure as to why this was changed.

PR #583, opened 2024-03-20
cache-addition-in-lora

SPOOKEXE:
This function causes HEAPS of delay for people with large numbers of lora models. Consider adding similar cache checks (and potential file-change checks) in these list available functions as they are incredibly important for people with large file numbers.

## Description

Improve the caching of models to reduce the latency when refreshing models. This is important for a pagination settings as well as the page needs to be refreshed and these particular functions are called.

## Checklist:

yes


SPOOKEXE:
13k lora models
image

without the cache value:
image

with the cache value:

PR #602, opened 2024-03-23
Add New Controlnet Nromal preprocesser Depth_fm

sdbds:
image

https://github.com/CompVis/depth-fm

Host depth_fm as a python package dependency

### DIFF
--- a/extensions-builtin/forge_legacy_preprocessors/annotator/depth_fm.py
+++ b/extensions-builtin/forge_legacy_preprocessors/annotator/depth_fm.py
@@ -1,0 +1,61 @@
+import os
+import torch
+import cv2
+import numpy as np
+import torch.nn.functional as F
+
+from depthfm import DepthFM
+from einops import rearrange
+from .util import load_model
+from .annotator_path import models_path
+from modules_forge.supported_preprocessor import Preprocessor, PreprocessorParameter
+from modules_forge.shared import preprocessor_dir, add_supported_preprocessor
+from modules_forge.forge_util import resize_image_with_pad
+
+
+class DepthFMDetector:
+    """https://github.com/CompVis/depth-fm"""
+
+    model_dir = os.path.join(models_path, "depth_fm")
+
+    def __init__(self, device: torch.device):
+        self.device = device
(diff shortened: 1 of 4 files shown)


huchenlei (commented):
We should probably stop adding new entries to the legacy preprocessor file. Can you make adjustment to follow the new preprocessor guideline?

https://github.com/lllyasviel/stable-diffusion-webui-forge#add-a-preprocessor

sdbds:
> We should probably stop adding new entries to the legacy preprocessor file. Can you make adjustment to follow the new preprocessor guideline?
> 
> https://github.com/lllyasviel/stable-diffusion-webui-forge#add-a-preprocessor

OK,  ill add it outside.

sdbds:
> We should probably stop adding new entries to the legacy preprocessor file. Can you make adjustment to follow the new preprocessor guideline?
> 
> https://github.com/lllyasviel/stable-diffusion-webui-forge#add-a-preprocessor

Because the author of the original depth_fm wrote model load_state_dict in the source code, it is difficult to load it using an external mod patcher, and after I tried it, I found that the part of the VRAM management is not easy to unload outside of the preprocessor, which is not effective, so I would suggest to put it in inside instead.

huchenlei:
> > We should probably stop adding new entries to the legacy preprocessor file. Can you make adjustment to follow the new preprocessor guideline?
> > https://github.com/lllyasviel/stable-diffusion-webui-forge#add-a-preprocessor
> 
> Because the author of the original depth_fm wrote model load_state_dict in the source code, it is difficult to load it using an external mod patcher, and after I tried it, I found that the part of the VRAM management is not easy to unload outside of the preprocessor, which is not effective, so I would suggest to put it in inside instead.

I think @lllyasviel implemented very aggressive memory management system that hijacks every torch.nn.Module creation and unload them when necessary. Manual unloading in preprocessor code is no longer necessary.

kagevazquez:
https://github.com/lpiccinelli-eth/UniDepth > This is more accurate than depth_anything and depth_fm.

huchenlei:
This can probably be hosted in a separate repo. Closing for now.

PR #1096, opened 2024-08-14
Fix validation error by updating SDModelItem config field to be optional

davidmartinrius:
Changed `config` field in `SDModelItem` model to `Optional[str]` with default value `None`.
Without this change, calling /sdapi/v1/sd-models will cause it to crash


ammduscias:
**Kek**

davidmartinrius:
> **Kek**

What do you mean @ammduscias ?

PR #1198, opened 2024-08-16
add/fix infotext loading for builtin extensions

DenOfEquity:
some simple changes, just busy-work really

for eight builtin extensions:
* save settings from process() not process_before_every_sampling() so they are written to params.txt too

for seven builtin extensions (latent modifier not changed):
* add/fix infotext loading (can't use .update with Gradio 4, this method works with v4 and v3)

for freeu:
* dropdown for presets for various sd versions, taken from /ChenyangSi/FreeU
* minor UI tweak, placed pairs of sliders into rows to use less space


lllyasviel:
i will wait several days before merge this and hope that someone can test if this break things

DenOfEquity:
I'll redo these individually, so it's not one big pile of changes.

PR #1208, opened 2024-08-16
Fix audio notification for Gradio 4

### LINKED ISSUE #1002: notification.mp3 didn't sound after generation
I placed a notification.mp3 in webui folder, it works in a1111 webui and gradio3 forge. but in forge 2.0.1 it didn't work anymore, there won't be a ring after generation.

chocolateboy:
The AUDIO element is now in the shadow DOM, so access it there to trigger playback. Closes #1002.

This can be reverted if the AUDIO element is exposed directly again via https://github.com/gradio-app/gradio/issues/8267.

### DIFF
--- a/javascript/notification.js
+++ b/javascript/notification.js
@@ -26,7 +26,7 @@ onAfterUiUpdate(function() {
     lastHeadImg = headImg;
 
     // play notification sound if available
-    const notificationAudio = gradioApp().querySelector('#audio_notification audio');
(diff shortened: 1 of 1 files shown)


WingeD123:
notification won't ring if Forge tab isn't the current active tab after generation complete.

if I switch to other tab after click generate button, it won't ring until i change back to Forge tab. even the generation is already complete. I am using EDGE

PR #1312, opened 2024-08-19
minor fixes related to Extras tab

DenOfEquity:
1. Send to Extras buttons work as far as sending the image, but do not switch to the Extras tab since the addition of Spaces. Single character change to correct the index used.
2. Extras batch processing fails due to missing attribute 'orig_name'. Corrected to use attribute 'name'. Tested: correctly processes and writes new images preserving the filename if that option is set. Directory processing works too.
3. Postprocessing parameters are written twice, once to PNG section 'postprocessing' and again to 'extras'. Removed one unnecessary line and updated image save to write new text to 'postprocessing' because that's a better name than 'extras'.

Bug not fixed:
single file extras processing doesn't copy generation parameters to the new image (https://github.com/lllyasviel/stable-diffusion-webui-forge/discussions/1204). Maybe something ForgeCanvas doesn't handle? Reverting to old method, using gradio Image, does work but I second-guessed that might be an unwanted change so didn't include it. Does ForgeCanvas have benefits for the Extras tab?
working : **modules/ui_postprocessing.py** : *line 16* :
`extras_image = gr.Image(label="Source", source="upload", interactive=True, type="pil", elem_id="extras_image", height=512)`

### DIFF
--- a/javascript/ui.js
+++ b/javascript/ui.js
@@ -82,7 +82,7 @@ function switch_to_inpaint_sketch() {
 }
 
 function switch_to_extras() {
-    gradioApp().querySelector('#tabs').querySelectorAll('button')[2].click();
+    gradioApp().querySelector('#tabs').querySelectorAll('button')[3].click();
 
     return Array.from(arguments);
 }
--- a/modules/postprocessing.py
+++ b/modules/postprocessing.py
@@ -93,7 +93,6 @@ def run_postprocessing(extras_mode, image, image_folder, input_dir, output_dir, 
 
             if opts.enable_pnginfo:
                 pp.image.info = existing_pnginfo
-                pp.image.info["postprocessing"] = infotext
 
             shared.state.assign_current_image(pp.image)
 
--- a/modules/postprocessing.py
+++ b/modules/postprocessing.py
@@ -24,7 +24,7 @@ def run_postprocessing(extras_mode, image, image_folder, input_dir, output_dir, 
                     fn = ''
                 else:
                     image = images.read(os.path.abspath(img.name))
-                    fn = os.path.splitext(img.orig_name)[0]
+                    fn = os.path.splitext(img.name)[0]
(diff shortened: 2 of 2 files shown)


lllyasviel:
I am not sure what happens since webui uses orig_name as https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/82a973c04367123ae98bd9abdf80d9eda9b910e2/modules/postprocessing.py#L24
and forge does not change related logics i think

also ForgeCanvas.background is logically equivalent to gr.Image and can receive any PIL/np image in any gradio calls is that helps

DenOfEquity:
I tracked down the **orig_name** discrepancy: It's a change to the output gradio File component - old version included **name** and **orig_name** (same filename based on original file, but different temporary gradio file paths). New version passes a much cleaner object which only has **name**.
Example from old (using A1111):
```
{'file': <_io.BufferedRandom name='Q:\\temp\\gradio\\tmp78ie4t2u'>, 'name': 'Q:\\temp\\gradio\\ec28ac724887fb25c999e6d209184e8683403fe0\\00026-1809548749.png', 'delete': False, '_closer': <tempfile._TemporaryFileCloser object at 0x000001E469507EB0>, 'orig_name': 'Q:\\temp\\gradio\\9bc3d2af2e350c917dec4efc5cb3dca957aa8bd0\\00026-1809548749.png'}
```
Same file with new:
```
{'name': 'Q:\\temp\\gradio\\ec28ac724887fb25c999e6d209184e8683403fe0\\00026-1809548749.png'}
```

---
The ForgeCanvas issue isn't touched by this PR. But previous generation parameters are not copied for me, or for the user that started the linked discussion. Instead of the 'info' dict entry containing `{'parameters': ... }`, it contains only `{'srgb': 0}`. 'mode' is changed to 'RGBA', maybe lost in conversion somewhere?

Issue #1326, opened 2024-08-20
sample_deis


kevin98761234:
AttributeError: module 'backend.modules.k_diffusion_extra' has no attribute 'sample_deis'
module 'backend.modules.k_diffusion_extra' has no attribute 'sample_deis'
Traceback (most recent call last):
  File "F:\StabilityMatrix\Data\Packages\Stable Diffusion WebUI Forge\modules_forge\main_thread.py", line 30, in work
    self.result = self.func(*self.args, **self.kwargs)
  File "F:\StabilityMatrix\Data\Packages\Stable Diffusion WebUI Forge\modules\img2img.py", line 248, in img2img_function
    processed = process_images(p)
  File "F:\StabilityMatrix\Data\Packages\Stable Diffusion WebUI Forge\modules\processing.py", line 814, in process_images
    res = process_images_inner(p)
  File "F:\StabilityMatrix\Data\Packages\Stable Diffusion WebUI Forge\modules\processing.py", line 880, in process_images_inner
    p.init(p.all_prompts, p.all_seeds, p.all_subseeds)
  File "F:\StabilityMatrix\Data\Packages\Stable Diffusion WebUI Forge\modules\processing.py", line 1598, in init
    self.sampler = sd_samplers.create_sampler(self.sampler_name, self.sd_model)
  File "F:\StabilityMatrix\Data\Packages\Stable Diffusion WebUI Forge\modules\sd_samplers.py", line 42, in create_sampler
    sampler = config.constructor(model)
  File "F:\StabilityMatrix\Data\Packages\Stable Diffusion WebUI Forge\modules_forge\alter_samplers.py", line 15, in constructor
    return AlterSampler(m, sampler_name)
  File "F:\StabilityMatrix\Data\Packages\Stable Diffusion WebUI Forge\modules_forge\alter_samplers.py", line 9, in __init__
    sampler_function = getattr(k_diffusion_extra, "sample_{}".format(sampler_name))
AttributeError: module 'backend.modules.k_diffusion_extra' has no attribute 'sample_deis'
module 'backend.modules.k_diffusion_extra' has no attribute 'sample_deis'

DenOfEquity:
Same for huen_pp2, ipndm, ipndm_v.
Sampling functions are looked for in `backend.modules.k_diffusion_extra` but were actually added in `k_diffusion.sampling`.
[...]

PR #1340, opened 2024-08-20
Fix for new samplers

DenOfEquity:
four new samplers added recently, failed on use: https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/1326

moved definitions - easiest fix.

Tested all OK.


Panchovix:
Okay perfect, did the testing and it works as expected. Thanks for the PR!

Issue #1343, opened 2024-08-20
XYZ Plot - Checkpoint Name gets ignored


RedDeltas:
I've found that since updating my Forge that the Checkpoint Name gets ignored in XYZ plots. The plot still renders the Checkpoint Name but it's clearly not loading / changing the checkpoint.

Version Info:
```
Python 3.10.12 (main, Mar 22 2024, 16:50:05) [GCC 11.4.0]
Version: f2.0.1v1.10.1-previous-363-g14ac95f9
Commit hash: 14ac95f9089a25171b30f3906f68613afa7abbfe
```
![Screenshot 2024-08-20 at 15 53 03](https://github.com/user-attachments/assets/8aadd337-606e-4a95-88ba-b416bd6f4947)
![Screenshot 2024-08-20 at 15 53 38](https://github.com/user-attachments/assets/48f744fb-c76a-4aba-a756-fe136601d0d6)

hopperelec:
Having the same issue here

Issue #1235, opened 2024-08-17
XYZ plot don't change FLUX chekpoints


Desm0nt:
If I select chekpoints as field for XYZ plot - it do nothin and generates all images with last (current) chekpoint, so we can't compare different Flux quants in a grid.

l0stl0rd:
same issue here

MBurneracc:
XYZ plots don't work for ANY checkpoints and haven't for like two weeks now, guy that reported it originally even got tired of waiting and found the cause themselves was Illya arbitrarily deleting the necessary code during the construction.  Fixing what they broke doesn't seem to be a priority.

Auryg:
> XYZ plots don't work for ANY checkpoints and haven't for like two weeks now, guy that reported it originally even got tired of waiting and found the cause themselves was Illya arbitrarily deleting the necessary code during the construction. Fixing what they broke doesn't seem to be a priority.

If it's that big of a deal for you revert to a previous commit.  lllya has been putting in a ton of work and it'll probably be easier for him to go back through and fix bugs once everything is in place rather than tiptoeing around the code.

kerizane:
Duplicate of https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/952

Asaghon:
Anyone know at which commit this still works? It's the main reason I use Forge for :p

Issue #952, opened 2024-08-07
X/Y/Z plot doesn't respect checkpoints from X Type, but use main selected checkpoint


shsmad:
On commit f743fbff83c7db0bf0957ab9718f8d42a47eb35e I found strange behavior, steps to reproduce:

1. enable X/Y/Z plot
2. select «Checkpoint name» for «X Type», select few checkpoints
3. enter description, sampling method & scheduler type, steps & cfg
4. hit generate

Expected results: different single images from different checkpoints & grid with different images on it

Got results: same single images, that are using only one checkpoint — main checkpoint selected in top dropdown, not those, which are entered in X Type

dxwarlock:
Came to report the same issue.
X/Y/Z does not use selected checkpoints

PixelClassicist:
Same issue here

kievhsiao:
Checkpoint name & VAE don't work.

shsmad:
I've found problem:

In old version e95333c5568b5039f7118d036c16afa9a2a8ad2c in `modules/processing.py` in `def process_images` was code, which respects checkpoints from xyz:

```py
        # if no checkpoint override or the override checkpoint can't be found, remove override entry and load opts checkpoint
        # and if after running refiner, the refiner model is not unloaded - webui swaps back to main model here, if model over is present it will be reloaded afterwards
        if sd_models.checkpoint_aliases.get(p.override_settings.get('sd_model_checkpoint')) is None:
            p.override_settings.pop('sd_model_checkpoint', None)
            sd_models.reload_model_weights()

        for k, v in p.override_settings.items():
            opts.set(k, v, is_api=True, run_callbacks=False)

            if k == 'sd_model_checkpoint':
                sd_models.reload_model_weights()

            if k == 'sd_vae':
                sd_vae.reload_vae_weights()

            res = process_images_inner(p)
```

so that `opts.set(k, v, is_api=True, run_callbacks=False)` set correct `sd_model_checkpoint` and `sd_vae` in `shared.opts` which are used later.

But in current version (00f1cd36) there is no such code in that file, only

```py
        res = process_images_inner(p)
```

Also in e95333c5568b5039f7118d036c16afa9a2a8ad2c in `modules/initialize_util.py` there are lines 

```py
    shared.opts.onchange("sd_model_checkpoint", wrap_queued_call(lambda: main_thread.run_and_wait_result(sd_models.reload_model_weights)), call=False)
[...]

PR #1386, opened 2024-08-21
Possible partial fix for #952, #1235, #1343

wcole3:
Initial attempt to fix xyz_grid script loading checkpoints.  Tried using the `onchange` in modules/initialize_util but got some odd lockups probably developer error.

This should minimize the reloading during a x/y/z grid run, but it does result in the first non-script run after running the xyz-grid to reload the model.

VAE changing between run still wip.

Ref issues: #952 #1235 #1343


lllyasviel:
My concern is that the name `refresh_loading_params` is too formal and looks like main entry for that functionality when it is not
maybe `refresh_loading_params_for_xyz_grid` or similar is a better name.
also maybe this func can be moved to xyz_grid file as they only serve xyz

wcole3:
Makes sense to me, refactored.

PR #1387, opened 2024-08-21
Variable-Strength StyleAlign

twarrendewit:
Added strength slider for the ratio of StyleAlign applied within a batch. I was finding that my images with StyleAlign replacing the entire attention stack was making output too same-y, especially when trying to generate one subject in different poses within a single run.

Prompt, using Dynamic Prompt for batches, not fixing seed:
_a plate of {salad | pizza | cheeseburgers | bones} with full table setting and a night harbor view_

**At Strength=0 (Equivalent to Off)**
align-00

**At Strength=0.2**
align-02

**At Strength=0.4**
align-04

**At Strength=0.6**
align-06

**At Strength
align-08
=0.8**

**At Strength=1.0 (Equivalent to Previous Check)**
align-10

### DIFF
--- a/extensions-builtin/sd_forge_stylealign/scripts/forge_stylealign.py
+++ b/extensions-builtin/sd_forge_stylealign/scripts/forge_stylealign.py
@@ -22,14 +22,15 @@ class StyleAlignForForge(scripts.Script):
     def ui(self, *args, **kwargs):
         with gr.Accordion(open=False, label=self.title()):
             shared_attention = gr.Checkbox(label='Share attention in batch', value=False)
+            strength = gr.Slider(label='Strength', minimum=0.0, maximum=1.0, value=1.0)
 
-        return [shared_attention]
+        return [shared_attention, strength]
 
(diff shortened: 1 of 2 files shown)


twarrendewit:
Adding one more example to show what I mean:

Dynamic Prompt: _a man in front of a brick wall in a business suit, {holding an umbrella | eating a snow cone}, {sitting on a bench | leaning against wall}_

**At Strength=0.00**
man-000


**At Strength=0.25**
man-025

**At Strength=0.50**
man-050

**At Strength=0.75**
man-075

**At Strength=1.00**
man-100

lllyasviel:
I think it would be better if you can only compute attention once when strength is exactly 0 or 1

twarrendewit:
Do you mean for efficiency (to avoid re-compute)?

lllyasviel:
`blended_attention = (1.0 - strength) * original_attention + strength * aligned_attention_result`

i mean you do not need to compute `original_attention`  or `aligned_attention_result` when strength is 0 or 1

twarrendewit:
Good catch, updated.

Issue #993, opened 2024-08-11
Important API Things you Already Know... txt2img, sd-models, etc


altoiddealer:
Your recent changes are absolutely marvelous, thank you so much for making Flux a viable option for low VRAM users.

I mainly use Stable Diffusion via API.

**I very much look forward to the essential API endpoints being restored:**

- txt2img
- sd-models
- img2img

I will be sure to close this issue when they are restored.

Thank you again!

Xerophayze:
I also use it primarily via API and would love to see an updated API documentation on any changes that were implemented in this.  Thank you for all your hard work on this.

altoiddealer:
> I also use it primarily via API and would love to see an updated API documentation on any changes that were implemented in this. Thank you for all your hard work on this.

Any changes to payload structures should be apparent when checking via http://127.0.0.1:7860/docs (address may be different to some).

GeminiSquishGames:
The move has been a bit wonky for api development. Currently broken for my application is the list of models, [/sdapi/v1/sd-models], a core need, but for now a hand imported list of my own servers checkpoints helps workaround for testing. /samplers and /schedulers works as expected. I guess pydantic is complaining about missing the model names or locations for validation. I'm also using a symlinked folder through Stability Matrix, unsure if that is related yet, but it was working before the big update times. Understandable with large updates.

I was going to post about this after I updated a while ago and it hadn't changed yet, but your issue was here so just adding my use case as an example rather than another issue. Hope that works okay and is of some benefit somehow.

dermesut:
there are some endpoints, that are broken.
looking at the api-link docs, they respond with "internal server error". checkpoints, extension list etc.

altoiddealer:
@lllyasviel  I recommend that you mention in **[Forge Status](https://github.com/lllyasviel/stable-diffusion-webui-forge?tab=readme-ov-file#forge-status)** that the API is `Broken, pending fix`

altoiddealer:
To elaborate:

I disabled all my extensions, go to `http://127.0.0.1:7860/docs`, expand `/sdapi/v1/txt2img` and click **Try it now** with default values and get this error:

```
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "C:\0_SD\stable-diffusion-webui-forge\venv\lib\site-packages\uvicorn\protocols\http\h11_impl.py", line 396, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
[...]

PR #1409, opened 2024-08-22
Fix 2 - FieldInfo object has no attribute type_

altoiddealer:
Fixes https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/993

I am pushing 2 different solutions for the  API error:
`AttributeError: 'FieldInfo' object has no attribute 'type_'`

The other PR: https://github.com/lllyasviel/stable-diffusion-webui-forge/pull/1406

### I do not know which PR is desirable in the new Forge setup.  Please merge one or the other, I trust you will know which is the ideal solution.

This PR resolves the error, **but does not** return the same values that previous WebUIs (A1111, old Forge, etc) would get from using `request.__fields__[field.api].type_`

- The `type_` attribute is now replaced with `annotation`
- However, `annotation` returns a typing construct (like `Optional[str]`)
- **This PR does NOT try to extract the "type" from the annotation**

### Screenshots with print statements:

**FORGE**

Forge

**A1111**

A111

### DIFF
--- a/modules/api/api.py
+++ b/modules/api/api.py
@@ -377,7 +377,7 @@ class Api:
                 return None
 
             if field.api in request.__fields__:
-                target_type = request.__fields__[field.api].type_
+                target_type = request.__fields__[field.api].annotation
             else:
                 target_type = type(field.component.value)
 


altoiddealer:
I originally submitted this with extra "catch up" commits.

Resubmitted without those.

PR #1453, opened 2024-08-23
Add API endpoint [GET] 'sd-text-encoders'

altoiddealer:
This API call should be useful soon, if not immediately.

This implementation is *kind of* sloppy, since **text_encoders** are not handled particularly graceful (yet).

As, far as I can tell, **text_encoders** are currently only fetched for the UI via **[refresh_models](https://github.com/lllyasviel/stable-diffusion-webui-forge/blob/df598c4df5ee8c6c6b12f1cd4c78043c6192cde2/modules_forge/main_entry.py#L136)** - lumped in with VAEs.

Meanwhile, **[get_sd_vaes](https://github.com/lllyasviel/stable-diffusion-webui-forge/blob/df598c4df5ee8c6c6b12f1cd4c78043c6192cde2/modules/api/api.py#L737)** endpoint has the dedicated [sdvae module](https://github.com/lllyasviel/stable-diffusion-webui-forge/blob/df598c4df5ee8c6c6b12f1cd4c78043c6192cde2/modules/sd_vae.py) to fetch from (which does not manage / return **text-encoders** at all).

I am a bit of an amateur, so I do not know how a **text_encoders module** could be fleshed out and put to good use.

In the meantime, this solution just uses **[find_files_with_extensions](https://github.com/lllyasviel/stable-diffusion-webui-forge/blob/f23ee63cb340504c15607cc68798cd28abb88037/modules_forge/main_entry.py#L126)** to populate the dictionary.

---

**Known bug:** Since `shared.cmd_opts.vae_dir` is a mixture of VAE/Text Encoders - by this API call including those directories, it will likely return normal VAE checkpoints if `--vae_dir` CMD Arg is passed.

```
        if isinstance(shared.cmd_opts.vae_dir, str): # TODO shared.cmd_opts.text_encoder_dir?
            text_encoder_paths.append(os.path.abspath(shared.cmd_opts.vae_dir))
```

**Possible solutions**
- Improve separation for **text_encoders** from **VAEs**
  - Add a new CMD Arg explicitly for `--text_encoders_dir` and use it in the API function.
  - Update **[refresh_models](https://github.com/lllyasviel/stable-diffusion-webui-forge/blob/df598c4df5ee8c6c6b12f1cd4c78043c6192cde2/modules_forge/main_entry.py#L136)** to collect both lists and merge them for the UI.

OR

- Just omit the 2 lines regarding `--var_dir` (Do not allow getting these text_encoders via API)

---

get_text_encoders

### DIFF
--- a/modules/api/api.py
+++ b/modules/api/api.py
@@ -224,6 +224,7 @@ class Api:
         self.add_api_route("/sdapi/v1/latent-upscale-modes", self.get_latent_upscale_modes, methods=["GET"], response_model=list[models.LatentUpscalerModeItem])
         self.add_api_route("/sdapi/v1/sd-models", self.get_sd_models, methods=["GET"], response_model=list[models.SDModelItem])
         self.add_api_route("/sdapi/v1/sd-vae", self.get_sd_vaes, methods=["GET"], response_model=list[models.SDVaeItem])
+        self.add_api_route("/sdapi/v1/sd-text-encoders", self.get_sd_text_encoders, methods=["GET"], response_model=list[models.SDTextEncoderItem])
         self.add_api_route("/sdapi/v1/hypernetworks", self.get_hypernetworks, methods=["GET"], response_model=list[models.HypernetworkItem])
         self.add_api_route("/sdapi/v1/face-restorers", self.get_face_restorers, methods=["GET"], response_model=list[models.FaceRestorerItem])
         self.add_api_route("/sdapi/v1/realesrgan-models", self.get_realesrgan_models, methods=["GET"], response_model=list[models.RealesrganItem])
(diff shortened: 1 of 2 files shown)


lllyasviel:
In Forge, vae and text encoder are same thing. you can put vae models in text encoder folder, or put text encoders in vae folder. they will all work and load properly

The detection is fully robust and will detect the type from state dict contents rather than folder. 

So the two folders are just for making users feel comfortable as user may want to put them in different folders

so perhaps you just need to query them together as "external modules"

altoiddealer:
> In Forge, vae and text encoder are same thing. you can put vae models in text encoder folder, or put text encoders in vae folder. they will all work and load properly
> 
> The detection is fully robust and will detect the type from state dict contents rather than folder.
> 
> So the two folders are just for making users feel comfortable as user may want to put them in different folders
> 
> so perhaps you just need to query them together as "external modules"

Well, here is another thing in the UI which only queries the "VAE" models, but not the "text encoders"

[XYZ - VAE](https://github.com/lllyasviel/stable-diffusion-webui-forge/blob/8ff0dc6fed9fa7ce0c15a9e9d423bf77149b8630/scripts/xyz_grid.py#L148)

XYZ Vae


IF ALL queries for **VAE** should now include **text encoders**, then I think the correct course of action would be:

- Revise **modules/sd_vae** to include `text_encoders` path

Well... that would do it, really.  Then the `get_sd_vae` API endpoint would return everything, and XYZ would return all models.

It's up to you whether we go nuts and rename everything with "VAE" / "Text Encoder" to some universal term

altoiddealer:
Alright... I see there is A LOT of code related to VAE / Text Encoders that should be consolidated.

Half of **modules/sd_vae** looks ancient compared to a few functions in **main_entry**


I am not personally ambitious enough to clean all of that up - so I will just push a commit that makes the existing API endpoint return all VAE / text encoders, and maybe even fix that XYZ thing...

Issue #1359, opened 2024-08-21
Recent commits broke merging


cluder:
trying to merge models fails now because of multiple methods being called which have been removed (pass) or commented out

https://github.com/lllyasviel/stable-diffusion-webui-forge/commit/c1b23bd49479c890255193c269857892c91376ec
https://github.com/lllyasviel/stable-diffusion-webui-forge/commit/bccf9fb23a2cbc9d42091a856b3d35b2c4414d51

1) sd_models.read_state_dict() has been removed (pass), but its still called during merging 
see extras.py: run_modelmerger()

2)   AttributeError: module 'modules.sd_models_config' has no attribute 'find_checkpoint_config_near_filename'
find_checkpoint_config_near_filename has been commented out in: bccf9fb23a2cbc9d42091a856b3d35b2c4414d51

the following changes seems to work for me, not sure if create_config() is needed at all after merging?

```$ git diff
diff --git a/modules/extras.py b/modules/extras.py
index 2a310ae3..af0e2cfd 100644
--- a/modules/extras.py
+++ b/modules/extras.py
@@ -179,7 +179,7 @@ def run_modelmerger(id_task, primary_model_name, secondary_model_name, tertiary_

     shared.state.textinfo = f"Loading {primary_model_info.filename}..."
     print(f"Loading {primary_model_info.filename}...")
-    theta_0 = sd_models.read_state_dict(primary_model_info.filename, map_location='cpu')
+    theta_0 = sd_models.load_torch_file(primary_model_info.filename)^M

     print("Merging...")
     shared.state.textinfo = 'Merging A and B'
@@ -222,7 +222,7 @@ def run_modelmerger(id_task, primary_model_name, secondary_model_name, tertiary_
     if bake_in_vae_filename is not None:
         print(f"Baking in VAE from {bake_in_vae_filename}")
         shared.state.textinfo = 'Baking in VAE'
-        vae_dict = sd_vae.load_vae_dict(bake_in_vae_filename, map_location='cpu')
+        vae_dict = sd_vae.load_torch_file(bake_in_vae_filename)^M

         for key in vae_dict.keys():
             theta_0_key = 'first_stage_model.' + key
@@ -321,7 +321,7 @@ def run_modelmerger(id_task, primary_model_name, secondary_model_name, tertiary_
     if created_model:
         created_model.calculate_shorthash()

-    create_config(output_modelname, config_source, primary_model_info, secondary_model_info, tertiary_model_info)
[...]

Issue #1095, opened 2024-08-14
Checkpoint Merger Error


AvaCadava:
*** Error loading/saving model file██████████████████████████████████████████████████| 140/140 [00:40<00:00,  3.71it/s]
    Traceback (most recent call last):
      File "F:\AI\StableDiffusion\WebUI-Forge\modules\ui_checkpoint_merger.py", line 21, in modelmerger
        results = extras.run_modelmerger(*args)
      File "F:\AI\StableDiffusion\WebUI-Forge\modules\extras.py", line 135, in run_modelmerger
        primary_model_info = sd_models.checkpoints_list[primary_model_name]
    KeyError: 'SD15_AvaCadava_v6_EP150.safetensors'

---

I tried different models / combinations and all of them failed with the same error. I get this error since I've updated to the new UI design with FLUX integration. Before the Update I had no problems to merge my models.

ayyRyuu:
I am getting the same error. I tried multiple fixes but none of them worked.

nbzncloud:
Same issue. I've always used Forge and I do a lot of checkpoint merges, but after the flux update I can't merge anything anymore. I'm having to use A1111 for this now.

raiken92:
I have the exact same issue too. Everything was working fine and then after I updated it stopped working. Are there any fixes yet?

cluder:
see also #1359, i found some problems, the KeyError seems to come from forge searching with 'SD15_AvaCadava_v6_EP150.safetensors'  but the list of checkpoints has the key like: 'SD15_AvaCadava_v6_EP150.safetensors [somehash]'  so it can't match

cluder:
should be fixed now

ayyRyuu:
> should be fixed now

Normally merging two checkpoints works fine but I'm getting "'NoneType' object has no attribute 'keys'" error when I try to add difference on three models

ssbugman:
> ge and I do a lot of checkpoint merges, but after the flux upd

Not fixed, now it become even worse and UI hangs on "waiting" after error.
Still "Error merging checkpoints: 'NoneType' object has no attribute 'keys'"

PR #1454, opened 2024-08-23
Fix Checkpoint Merging #1359,#1095

cluder:
see also #1359

- checkpoint_list[] contains the CheckpointInfo.title which is "checkpointname.safetensor [hash]" when a checkpoint is selected to be loaded during merge, we try to match it with just "checkpointname.safetensor". -> use checkpoint_aliases[] which already contains the checkpoint key in all possible variants.
- replaced removed sd_models.read_state_dict() with sd_models.load_torch_file()
- replaced removed sd_vae.load_vae_dict() with sd_vae.load_torch_file()
- uncommented create_config() for now, since it calls a removed method: sd_models_config.find_checkpoint_config_near_filename()

not sure what to do with create_config() it it even needed?

### DIFF
--- a/modules/extras.py
+++ b/modules/extras.py
@@ -132,17 +132,17 @@ def run_modelmerger(id_task, primary_model_name, secondary_model_name, tertiary_
     if not primary_model_name:
         return fail("Failed: Merging requires a primary model.")
 
-    primary_model_info = sd_models.checkpoints_list[primary_model_name]
+    primary_model_info = sd_models.checkpoint_aliases[primary_model_name]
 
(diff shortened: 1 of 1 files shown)


ssbugman:
Now it even more broken, model A dropbox becomes empty after pushing Merge button and before process starts so it fails.
Same error still appears.

cluder:
it's hard to have it 'more broken' when it didn't work at all before, can you post your commit and error message?
`Commit hash: d339600181252aafbe12dc06f1b4fa7f953fc6f3`
For me it does work.
image

Issue #1455, opened 2024-08-23
Settings in UI issue


LesPles:
With a clean install, I've noticed a new issue recently when I configure "Settings in UI" and add "face_restoration_model", save the changes, restart the UI, after restart I generate an image with a face with a static seed, then check the "Restore faces" box in the Txt2Img UI tab, and re-generate, the face does NOT go through the restoration process.

The only way you can force face restoration in recent Forge commits is on the Settings page.

This is a change of behavior as this is working fine in an old commit:
version: [f0.0.17v1.8.0rc-latest-278-gbfee03d8](https://github.com/lllyasviel/stable-diffusion-webui-forge/commit/bfee03d8d9415a925616f40ede030fe7a51cbcfd)

DenOfEquity:
Confirmed by my own tests. Not sure exactly what's going wrong yet.
If you add to Quicksettings (Settings > User interface), that does work, but will be at top of page instead.

PR #1465, opened 2024-08-23
restore handling of override_settings

DenOfEquity:
I've prevented overrides for checkpoint and vae, guessing that was why this was originally removed (testing suggests that checkpoint/vae overrides are currently not set from infotext anyway). Also noted that Quicksettings overrides can still apply.

https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/1455


lllyasviel:
This is adding another layer of complexity to maintain and we are going to set it on hold until every other things are almost fixed

DenOfEquity:
closing as superseded by #2027

PR #1482, opened 2024-08-24
more t5 quants

lllyasviel:
Although I still find neither speed advantage nor quality advantage for this, based on my highest love to Forge users, I spend 20 minutes to add those code, and I hope this is the last time that people randomly invent  new state dict key formats to waste time of other developers.

### DIFF
--- a/backend/loader.py
+++ b/backend/loader.py
@@ -10,7 +10,7 @@ from diffusers import DiffusionPipeline
 from transformers import modeling_utils
 
 from backend import memory_management
-from backend.utils import read_arbitrary_config, load_torch_file
+from backend.utils import read_arbitrary_config, load_torch_file, beautiful_print_gguf_state_dict_statics
 from backend.state_dict import try_filter_state_dict, load_state_dict
 from backend.operations import using_forge_operations
 from backend.nn.vae import IntegratedAutoencoderKL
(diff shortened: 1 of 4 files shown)


pflky:
It was worth it, because GGUF Q4 checkpoint + GGUF Q8 FP16 Text encoder combined virtually fits into 12GB VRAM now. Same if we had just the UNET from BNB NF4, that would fit in 12GB with it too. And the FP16 text encoder does make a big difference compared to FP8. You can get a lot more garbled and glitchy details with the FP8 text encoder.

danilomaiaweb:
> It was worth it, because GGUF Q4 checkpoint + GGUF Q8 FP16 Text encoder combined virtually fits into 12GB VRAM now. Same with NF4, that now fits in 12GB with it too. And the FP16 text encoder does make a big difference compared to FP8. You can get a lot more garbled and glitchy details with the FP8 text encoder.
> 
> Since we know that Q8 checkpoints deliver virtually the same output as the full sized checkpoint, we can assume that the Q8 text encoder delivers about the same as the full sized FP16 version, at half the size.
> 
> And for people with 24GB VRAM, they can now load both the Q8 checkpoint and Q8 encoder and even potentially have room for doing batches. If they do the NF4 or Q4 checkpoint and Q8 encoder, they basically have enough room in VRAM for controlnet models too, and also room for LORAs for people with lower VRAM.
> 
> As far as I understand, Forge does memory management that swaps things anyways so that only one is loaded at a time, but even with 12GB VRAM, I'm in the negative of VRAM availability by 700MB with anything higher than the Q5_KS model.
> 
> From my experience, while you can get really good images out of all models, anything below Q6 does have more visual issues, especially in the fine details where you can get choppy over sharpened looking artifacts, and bokeh backgrounds that end up with weird texture patterns in them. 1024x1024 can end up looking like if 768x768 was upscaled and over sharpened.
> 
> And if we could apply all this retroactively to SD 1.5 and SDXL with NF4 or GGUF conversions, we can really shrink things down. I don't know how it all works exactly, but if NF4 could be compressed with GGUF, a Q8 GGUF of NF4 should deliver about the same output as base NF4, while being even smaller. This could potentially enable the use of Flux + Controlnet models and even LORAs in 8-12GB of VRAM at once.




Sorry for the intrusion, but please, if possible, answer this question for me:

In your opinion, what is the best GGUF alternative for those who are using the bnb-nf4 version and only have an RTX 8GB+32RAM. 

**GGUF Q2.K Q3.KS Q4/Q4.1/Q4.KS Q5/Q5.1/Q5.KS Q6.K Q8 ???**

1. Which of these versions below do you think is a good option?
2. And what would be a basic configuration of steps and samplers, example: 10 steps, euler , normal?

Thank in advance here from Brazil !

pflky:
Likely Q4 unless you want to spill over into using system RAM, it's the only one that is less than 8GB. The text encoder's you can find in GGUF format should help too, and even the Q8 of those is pretty much. So Q4 or NF4 checkpoint with Q8 text encoder should work fine, since it's basically the same as using the full FP16 text encoder, except it's 5GB instead of 9GB.

danilomaiaweb:
Thank for your reply mate, please see this...

Output my console...
image

See the prompt, model, sampler, steps...
image

See the generated image in 15 steps. Time taken: 1 min. 27.6 sec.




I think that with only 15 steps the result does not have as much definition in the details, but if I increase it, the time to generate the image will also be longer, as I am impatient and want to see the result as quickly as possible, I only use 15 steps with this Q8 model

In another merged dev+schnell model (https://civitai.com/models/630820?modelVersionId=735521) I use only 6 steps and the generation time is only 24s

If you have any comments to make, feel free
And I thank you in advance for your time and comment!

danilomaiaweb:
Now with other model ArtFusion Flux 4-steps (https://civitai.com/models/641214/artfusion-flux-4-steps)...

Params:
steps 6, sampler Euler, Schedule type Normal, cfgScale	1, seed 807663290, size	584x1040

See the image result:


In this model the characters are dressed in their original clothes, while in the other model Q8, they change clothes, wearing gangster suits :)

Issue #922, opened 2024-08-02
api - get extension list not working


dermesut:
trying out http://127.0.0.1:7860/docs#/default/get_extensions_list_sdapi_v1_extensions_get with no extensions installed, i get a respsonse with an empty list - as expected.

when i install a first extension (say, ultimate sd upscale), go to the api-doc and try to execute:
http://127.0.0.1:7860/docs#/default/get_extensions_list_sdapi_v1_extensions_get

i get:
Undocumented
Error: Internal Server Error
Response body:
Internal Server Error

in consequence, the terminal shows these errors:
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "E:\ai\_gh_repos\webui_forge_cu121_torch21_EXP\system\python\lib\site-packages\uvicorn\protocols\http\h11_impl.py", line 404, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "E:\ai\_gh_repos\webui_forge_cu121_torch21_EXP\system\python\lib\site-packages\uvicorn\middleware\proxy_headers.py", line 84, in __call__
    return await self.app(scope, receive, send)
  File "E:\ai\_gh_repos\webui_forge_cu121_torch21_EXP\system\python\lib\site-packages\fastapi\applications.py", line 1106, in __call__
    await super().__call__(scope, receive, send)
  File "E:\ai\_gh_repos\webui_forge_cu121_torch21_EXP\system\python\lib\site-packages\starlette\applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "E:\ai\_gh_repos\webui_forge_cu121_torch21_EXP\system\python\lib\site-packages\starlette\middleware\errors.py", line 184, in __call__
    raise exc
  File "E:\ai\_gh_repos\webui_forge_cu121_torch21_EXP\system\python\lib\site-packages\starlette\middleware\errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "E:\ai\_gh_repos\webui_forge_cu121_torch21_EXP\system\python\lib\site-packages\gradio\route_utils.py", line 724, in __call__
    await self.app(scope, receive, send)
  File "E:\ai\_gh_repos\webui_forge_cu121_torch21_EXP\system\python\lib\site-packages\starlette\middleware\exceptions.py", line 79, in __call__
    raise exc
[...]

PR #1484, opened 2024-08-25
Correct commit_date type to 'int'

altoiddealer:
Resolves https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/922

@dermesut said it worked if using `int` instead of `str`, so I added debug printing to confirm.

Yes, `commit_date` is definitely an `int`

Screenshot 2024-08-24 211244

Error resolved by changing `commit_date: str` to `commit_date: int`

Screenshot 2024-08-24 211617


Panchovix:
Thanks for the PR!

Issue #1485, opened 2024-08-25
QoL issue with PNG info tab


sangoi-exe:
In the original WebUI, if there was already an image in the PNG info tab, I could just drag another one on top of it, and it would be replaced. Now, every time I want to check an image, I have to remove the current one first. 😕

PR #1489, opened 2024-08-25
allow repeated drag-drop into PNGInfo tab

DenOfEquity:
removes old special case for pnginfo_image element

fixes https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/1485


Panchovix:
Thanks for the PR! Confirmed it works as expected

PR #1507, opened 2024-08-26
add startup argument for text encoders location

DenOfEquity:
also tweaked description for clip location argument, for clarity

used same method as --vae-dir


lllyasviel:
you can merge all things like these without my review

PR #1529, opened 2024-08-26
Revert to proper hashing implementation in hashes.py

Melyns:
Restored the original functionality in hashes.py by removing the forge_fake_calculate_sha256 function. This ensures that models are now hashed correctly, and images are saved with accurate model hash metadata. https://github.com/lllyasviel/stable-diffusion-webui-forge/discussions/1477 https://github.com/lllyasviel/stable-diffusion-webui-forge/discussions/1242


lllyasviel:
The hash in a1111 has severe performance problems on some devices when model is larger than about 10GB and this is why Forge put them on hold.

Melyns:
I know your change was there for a reason but I wondered maybe you forgot to revert it. Thanks for the explanation.

bkosowski:
@lllyasviel @Melyns Suggestion: as this feature is both quite useful and somewhat performance-deteriorating on some systems, it would make sense to conditionally enable it via a command line switch. For example, by default it could be enabled, but it could be disabled via the cmd switch. It could also be properly documented in README or on Wiki page.

PR #1538, opened 2024-08-27
Reimplement Brush Color Settings

Haoming02:
- Right now, the 3 brush color options in **Settings** -> **img2img** are not used at all. So I add them back to the `ForgeCanvas` to make them function again.
- Also added a new option to toggle between the high-contrast brush and the color brush for the **Inpaint** mode, as some users may feel uncomfortable with the checked pattern ( #1528 )


DenOfEquity:
Nice one, works well.
I'd argue for including `scribble_color_fixed=True, scribble_alpha_fixed=True, scribble_softness_fixed=True` on the Inpaint canvas.

Haoming02:
> `scribble_color_fixed=True`, `scribble_alpha_fixed=True`, `scribble_softness_fixed=True`

imho it is better to let the user change those parameters. Disabling those kinda defeats the point of using `ForgeCanvas` ...

> now you never worry about brush color being invisible when it is similar to image color ( #853 )

One of the advantages of `ForgeCanvas` is that you can draw a mask on a white/bright image, without losing where the mask is.

Issue #1323, opened 2024-08-20
Sampler and CFG defaults are not being saved as defaults. 


andy8992:
I use distilled models/lora so I am forced to change these each time. I have gone and updated the defaults dozens of times and it always defaults to 7 and DPM(++afiafdifpa) or whatever it's called. 

In the defaults section in settings, it says it's making the change and shows what the settings are but they are never respected. 

IMG2IMG tab is also the same.

serick4126:
I’ve noticed that the default settings cannot be modified, and after investigating, I found that the configuration items named forge_preset are hard-coded in modules_forge/main_entry.py, making them difficult to change. 

While it would be possible to move these configuration items to a file like forge_preset.json, given the current situation where features are being frequently added, the existing setup might be more efficient.
(Additionally, the current form allows for the easy distribution of recommended settings by lllyasviel)

I’m interested to know if there are any plans to make the forge_preset more customisable in the future.

andy8992:
Though you are able to save other settings that wouldn't be compatible with 7cfg and the given sampler. It doesn't allow me to set 4 steps as default.

Not to mention it DOES say it is changing the defaults when you go to save them. It just doesn't.

Reavski-GH:
Having the same issue, made worse because my go-to models require specific settings/samplers. Having the UI claim to save the defaults, only for it change the settings to something incompatible when I wasn't looking must have shaved off at least a few years of my life before I realised what was happening.

Having a button/option to automatically override the baked-in settings with those from the edited user-config.json (they actually do get saved there) would be an excellent solution. As it is, starting the UI, reloading it or using the UI switcher ignores user-config wholesale.

coghazard:
Same issue.

MartyMcStarfox:
Very annoying :/

serick4126:
PR[#1541](https://github.com/lllyasviel/stable-diffusion-webui-forge/pull/1541) has been merged so that the “All” preset now applies the settings saved in ui-config.json.

PR #1541, opened 2024-08-27
Apply settings in ui-config.json to "All" presets

serick4126:
Added a preset “ui-config” that loads settings stored in ui-config.json.
This allows users to specify and save default values in the traditional way.

The settings to be loaded are the following items
- txt2img/Width/value
- txt2img/Height/value
- txt2img/CFG Scale/value
- txt2img/Distilled CFG Scale/value
- customscript/sampler.py/txt2img/Sampling method/value
- customscript/sampler.py/txt2img/Schedule type/value
- img2img/Width/value
- img2img/Height/value
- img2img/CFG Scale/value
- img2img/Distilled CFG Scale/value
- customscript/sampler.py/img2img/Sampling method/value
- customscript/sampler.py/img2img/Schedule type/value

This modification is related to the following issues
https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/1323

### DIFF
--- a/modules_forge/main_entry.py
+++ b/modules_forge/main_entry.py
@@ -62,7 +62,7 @@ def make_checkpoint_manager_ui():
         if len(sd_models.checkpoints_list) > 0:
             shared.opts.set('sd_model_checkpoint', next(iter(sd_models.checkpoints_list.values())).name)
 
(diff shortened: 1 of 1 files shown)


lllyasviel:
can we directly modify the "All" option rather than having a new option?

serick4126:
Of course it is possible. It is more simple that way.
The pull request has been changed from adding presets to applying ui-config.json settings to ALL presets.

Issue #1586, opened 2024-08-29
Checkpoint dropdown in Gradio 4 - Too small, left-aligned and filter area is buggy


MNeMoNiCuZ:
When switching to the new version of forge, the checkpoint list has a couple of issues for me when I updated my Forge to the latest version. From this version:
![image](https://github.com/user-attachments/assets/9908495c-b6b7-4ce4-b5a4-11e913a1d542)

To this version:
![image](https://github.com/user-attachments/assets/68407b38-2c10-4640-ad54-da3e75e07e14)

### Problem 1 - Dropdown width
The dropdown is slightly smaller than it was before. I understand the reason, and it is fine, with a few caveats.
When the checkpoint dropdown is clicked, the new width is limiting the visibility of long paths. I know there's an option to ignore paths and just show the checkpoint names, but this is not a solution. The paths are there for a reason, to sort the checkpoints properly. I wish to see the paths, but also see the name of the file. With new Forge I cannot.

How it is now:
![image](https://github.com/user-attachments/assets/262f9f37-fe33-4b22-b932-222ad68654fc)

How it was before:
![image](https://github.com/user-attachments/assets/19bc6d17-cfd1-411f-8bf8-7fb921eff866)

The key issue here is that it no longor shows the full width when the dropdown is clicked. I'm fine with the new shorter width when it's collapsed, for UI-purposes. But not at the cost of readability when clicking the menu.

---
### Problem 2 - Text alignment
There has also been a change in the maximum width of the dropdown menu when it's expanded. This makes the list unreadable when you have long paths. I have extra folders added to the start due to how I have my checkpoint SSDs set up with symlinks and such.

How it is now:
![image](https://github.com/user-attachments/assets/6bcf3a76-29bd-42c6-9411-1c25b4d19020)

As you can see, not seeing the filename is troublesome. Ideally this is not needed if we solve Problem 1, but if not, we should right-align this list to make it readable.

---
### Problem 3 - Filter typing
When typing in the dropdown, it used to be that the text you write is shown in the entire dropdown view, this means that you can read what you type.

How it used to be:
![image](https://github.com/user-attachments/assets/aba3c95b-15b2-4f01-b770-3e933d6af717)

How it is now:
![image](https://github.com/user-attachments/assets/a11bf1b7-f005-481c-9a54-0dbae47d23dd)

But as it is right now, I only get the typing area overlapping with the down-arrow on the right side of the drop-down.
[...]

PR #1588, opened 2024-08-29
Adjust Checkpoint input style

wcole3:
Slight adjustments to the Checkpoint dropdown in response to #1586 .

I would appreciate feedback from UX/UI folks, I think that these changes allow the filtering input to not overlap the dropdown arrow, but the right-align might be too much.

Current Filtering
image

Updated Filtering (See Below)

Current Long Checkpoint Listing
image

Updated Long Checkpoint Listing (See below)

Current Short Checkpoint Listing
image

Updated Short Checkpoint Listing (See below)

Additionally, not sure what the best way to address the ul.options formatting, I think adjusting them to right aligned would address the issue, but unsure if that is desired.

### DIFF
--- a/style.css
+++ b/style.css
@@ -443,6 +443,11 @@ div.toprow-compact-tools{
     min-width: 20em !important;
 }
 
+#quicksettings > div.model_selection input{
+    margin-right: 2.5em !important;
+    text-align: right !important;
+}
+
 #quicksettings .subdued{
     display: block;
     margin-left: auto;
--- a/style.css
+++ b/style.css
@@ -444,8 +444,8 @@ div.toprow-compact-tools{
 }
 
 #quicksettings > div.model_selection input{
-    margin-right: 2.5em !important;
-    text-align: right !important;
+    width: 85% !important; /* Ensure the input field is never chopped off */
+    margin-left: 2% !important; /* override the margin set for subdued */
 }
 
 #quicksettings .subdued{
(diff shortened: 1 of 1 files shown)


MNeMoNiCuZ:
Can you show a version where you expand the menu, and it shows a too long path? Does it show the whole thing? Or is this what you weren't able to figure out yet?

---

### Feedback:
I feel like the right-alignment is noone's friend. It just adds too much noise to one area and it's harder to parse.

wcole3:
That's what I couldn't figure out
image

wcole3:
Looking at it a bit more, I think you're right that the right align is not great so I removed it.  Also fixed an issue causing the input area to only be a fraction of the total area.

Updated LaF
image
image

The behavior of the dropdown for longer paths is still very wonky, I could get it to overflow, but I couldn't find a way to make the size match the size of the longest item. 
image
image

I did not add the above css changes, and I'll wait to see if others have ideas or if the scenario should be addressed another way.

MNeMoNiCuZ:
Nice!
Left-aligned is an improvement, and even the text going outside is an improvement in readability even if not perfect.

wcole3:
So I was able to find the right syntax to cap the length of the window at the max element.  I added that to the PR, but if its not desired I can back it out.

image

image

wcole3:
Can also cap the max width so that the popup doesn't go too crazy:

image

MNeMoNiCuZ:
I prefer the maximum length one myself but I'm open to other preferences.
Maybe it should be limited to the total width of the page, so it doesn't go to second monitor or outside of view.

Good job with this quick fix!

PR #1589, opened 2024-08-29
{prompt} in Lora's Activation text

Juqowel:
Same function as in Styles. Puts activation text before and after {prompt}.


lllyasviel:
is this your codes or some a1111 stuffs?

Juqowel:
I did it.

lllyasviel:
can you explain what is this?

Juqowel:
Not sure if this explanation is what needs.

Lora card:



**Result** of click on lora card with `Main text,` prompt:



Second click removes it.
Works only in positive prompt.

Also, fixed some problems now.

Panchovix:
Can you explain in more detail what this change does? Does it change a function how it works now, etc?

Juqowel:
> Can you explain in more detail what this change does? 

Just what I have already described, nothing more. Works only in positive and only if it is written once.

> Does it change a function how it works now, etc?

If you don't write a `{prompt}`, everything works as it does now.

PR #1639, opened 2024-08-31
Fix model loading during Checkpoint Merging #1359,#1095 

cluder:
replaced two instances of   
`sd_models.read_state_dict(secondary_model_info.filename, map_location='cpu')`
with 
`sd_models.load_torch_file(tertiary_model_info.filename)
`

before: no actual merging has been done, since the 2nd model was not loaded
image

after:
image


cluder:
@lllyasviel  this one seems pretty important as merging silently returns the first model

cluder:
also maybe read_state_dict() should be completely removed to avoid those problems?
`def read_state_dict(checkpoint_file, print_global_state=False, map_location=None):
    pass
`

PR #1647, opened 2024-09-01
reactivated the old hashing that work for civitai

trunksn1:
This pull request reactivates existing code in _hashes.py that was previously commented out or disabled.

### Issue
The current hashing function does not correctly identify resources on Civitai. As a result, checkpoints (both Stable Diffusion and Flux) and Loras (only Stable Diffusion) are not recognized.

### Fix: 
By reactivating the original hashing function, the recognition of these resources is restored:

1. Flux Checkpoints: Now recognized correctly.

2. Stable Diffusion Models: Both checkpoints and Loras are correctly recognized.

### Impact
**Before the Fix:**
Using the current fake hashing function, no resources are recognized.
fake_hash

**After the Fix:**

1. The Flux checkpoint is recognized, though Loras for Flux models are still not recognized.

good old hash

2. For Stable Diffusion models, both checkpoints and Loras are correctly recognized.
pony

### DIFF
--- a/modules/hashes.py
+++ b/modules/hashes.py
@@ -20,7 +20,8 @@ def calculate_sha256_real(filename):
 
 
 def calculate_sha256(filename):
-    return forge_fake_calculate_sha256(filename)
+    print("calculating real hash")
+    return calculate_sha256_real(filename)
 
 
 def forge_fake_calculate_sha256(filename):
@@ -60,7 +61,7 @@ def sha256(filename, title, use_addnet_hash=False):
         return None
 
     print(f"Calculating sha256 for {filename}: ", end='')
-    sha256_value = forge_fake_calculate_sha256(filename)
+    sha256_value = calculate_sha256_real(filename)
     print(f"{sha256_value}")
 
     hashes[title] = {
--- a/modules/hashes.py
+++ b/modules/hashes.py
@@ -20,7 +20,7 @@ def calculate_sha256_real(filename):
 
 
 def calculate_sha256(filename):
-    print("calculating real hash")
+    print("Calculating real hash: ", filename)
     return calculate_sha256_real(filename)
 
 
@@ -60,7 +60,7 @@ def sha256(filename, title, use_addnet_hash=False):
     if shared.cmd_opts.no_hashing:
         return None
 
-    print(f"Calculating sha256 for {filename}: ", end='')
+    print(f"Calculating sha256 for {filename}: ", end='', flush=True)
     sha256_value = calculate_sha256_real(filename)
     print(f"{sha256_value}")
 


devilkad:
thanks i tested it like this: 
1- deleting **hashes folder** inside **cache folder** 
2- replaced **hashes.py** inside **modules Folder**
3- inside webui go to **settings> actions tab> calculate hash for all check points**

devilkad:
and there is another fix you can setup civitai extension if you don't wanna bother yourself a lot coding
https://github.com/civitai/sd_civitai_extension

RalFingerLP:
> thanks i tested it like this: 1- deleting **hashes folder** inside **cache folder** 2- replaced **hashes.py** inside **modules Folder** 3- inside webui go to **settings> actions> calculate hash for all checkpoint**

thanks!

devilkad:
> > and there is another fix more user friendly 😊 you can setup civitai extension if you don't wanna bother yourself a lot coding https://github.com/civitai/sd_civitai_extension
> 
> Afaik that doesn't help since it uses the hash forge generates, it doesn't work for me at least..

for me both methods worked and civitai extension took like 10 minutes to load all the correct hashes from civitai and worked

UnconnectedBedna:
> > > and there is another fix more user friendly 😊 you can setup civitai extension if you don't wanna bother yourself a lot coding https://github.com/civitai/sd_civitai_extension
> > 
> > 
> > Afaik that doesn't help since it uses the hash forge generates, it doesn't work for me at least..
> 
> for me both methods worked and civitai extension took like 10 minutes to load all the correct hashes from civitai and worked

I'm no expert, but I am pretty sure the civitai extension does not change the hashes in the local database, only add meta info and dowload pictures etc.

I have used that plugin since forever when I started using forge, still have the incorrect hashes.

Applying this patch worked for me and civitai now recognizes my models from the meta tags after recalculating them in settings > actions > calculate hashes....
Thank you @devilkad for debugging and finding a fix for this issue.

SpaceShogox:
I wonder if there is any reason this change in the new forge isn't reverted or fixed yet. Hashing is a fundamental part of any software and it does break the use of Forge. @lllyasviel please take time to review the change needed. I don't know what this affects in the code but if not fixed it does affect the working of plugins.

IrineSistiana:
bump

I saw users complaining about strange hashing issues in a web community today, so I thought of this PR again.

Fake hash not just breaking Civitai but also many other AI image viewing tools that relay on hash to sort images. Silently using fake hashes can cause great confusion to end users who are not aware of the situation. (It's not just lack of hash in the metadata, it's a fake hash... "Is my checkpoint/SSD/memory broken?", "Why the hash changed?"...)

Normally hashing a checkpoint takes 5~20s depends what checkpoint and computer. It's a bit long but is totally fine I guess. There is a hash cache, so as long as the checkpoint path does not change, the hash will just calculate once. 

I'm using forge on local computer, not sure about other env, like on Colab. 

My suggestion is to add a new startup parameter, such as "--fake-hash", to "explicitly" disable hashing. If users know that they do not need hash and want to speed up the first time running, they can add this startup parameter.

Cranbert:
It's utterly baffling that this is still an issue. Saving a few seconds the first time you use a model is so insignificant compared to the annoyance it causes.

lBleyl:
This is very necessary, the amount of trouble this fake hash has bothered me is not written. This should at least be a toggle rather than simply removing the true hash.

DenOfEquity:
I thought about putting this change behind a setting, but there's already a cmd-opt for no-hashing.
Edited to flush the print before hashing, so user can have some idea that work is being done.

RalFingerLP:
Thank you for the merge!

Styl2sh:
lycoris doesn't seem to be affected by the fix, they still get fake hashes

Issue #1641, opened 2024-08-31
Many SDXL controlnet models that work in A1111 still do not work in Forge


pflky:
**Issue:**
---
RuntimeError: Input type (struct c10::Half) and bias type (float) should be the same
---

**Notes:**
Has been a prevalent issue for a long time. Only present in Forge controlnet, not A1111 controlnet. Controlnet models simply error out almost immediately when trying to run them. I previously opened this as a ticket before but it was closed with the reason of "unable to reproduce", so I am trying again to see if the problem can be remedied. I have found that the startup flag of "All in FP32" does seem to allow the models to work, but that is not really a solution since running everything in FP32 is not desirable for speed and memory usage. A possible factor in the issue may be SDXL models that have built-in LORAs, but unsure. Have tried many different remedies with no success. 

Here's a link to some example models that show this issue: 

[bdsqlsz depth]
https://civitai.com/models/136070?modelVersionId=267507
[bdsqlsz normal]
https://civitai.com/models/136070?modelVersionId=267533

Affects many other models as well, in fact, it affects most controlnet models in the CNXL page on Civit. Here is a link to the SDXL collection of controlnet models on Civit:

https://civitai.com/models/136070/controlnetxl-cnxl

---

**Full console log:**

[Unload] Trying to free all memory for cuda:0 with 0 models keep loaded ... Done.
StateDict Keys: {'unet': 1680, 'vae': 248, 'text_encoder': 197, 'text_encoder_2': 518, 'ignore': 0}
Working with z of shape (1, 4, 32, 32) = 4096 dimensions.
K-Model Created: {'storage_dtype': torch.float16, 'computation_dtype': torch.float16}
Model loaded in 0.9s (unload existing model: 0.1s, forge model load: 0.8s).
2024-08-31 16:29:09,928 - ControlNet - INFO - ControlNet Input Mode: InputMode.SIMPLE
2024-08-31 16:29:09,938 - ControlNet - INFO - Using preprocessor: normalbae
2024-08-31 16:29:09,938 - ControlNet - INFO - preprocessor resolution = 512
Loading base model ()...Done.
Removing last two layers (global_pool & classifier).
[Unload] Trying to free 2640.15 MB for cuda:0 with 0 models keep loaded ... Done.
[...]

Issue #1488, opened 2024-08-25
Controlnet CNET and CLite models are completely broken on SDXL


Allwhey:
Tested with Pony Diffusion, I used both a Controllite SDXL Canny Model and a Controlnet SDXL line art model, neither of them were able to generate. I am using the latest version of the repo.

Example of a CN Lineart Model:
`2024-08-25 01:44:15,628 - ControlNet - INFO - Preview Resolution = 1024
2024-08-25 01:44:22,350 - ControlNet - INFO - Preview Resolution = 1024
2024-08-25 01:44:24,826 - ControlNet - INFO - ControlNet Input Mode: InputMode.SIMPLE
2024-08-25 01:44:25,727 - ControlNet - INFO - Using preprocessor: lineart_standard (from white bg & black line)
2024-08-25 01:44:25,727 - ControlNet - INFO - preprocessor resolution = 1024
2024-08-25 01:44:26,031 - ControlNet - INFO - Current ControlNet ControlLLLitePatcher: U:\SD\stable-diffusion-automatic1111-webui\models\ControlNet\controlnetxlCNXL_bdsqlszLineart.safetensors
INFO:sd_dynamic_prompts.dynamic_prompting:Prompt matrix will create 8 images in a total of 1 batches.
To load target model JointTextEncoder
Begin to load 1 model
[Unload] Trying to free 5650.87 MB for cuda:0 with 0 models keep loaded ...
[Unload] Current free memory is 3839.57 MB ...
[Unload] Unload model KModel
[Memory Management] Current Free GPU Memory: 9544.89 MB
[Memory Management] Required Model Memory: 1752.98 MB
[Memory Management] Required Inference Memory: 3372.00 MB
[Memory Management] Estimated Remaining GPU Memory: 4419.91 MB
Moving model(s) has taken 6.16 seconds
[Unload] Trying to free 3372.00 MB for cuda:0 with 1 models keep loaded ...
[Unload] Current free memory is 7785.09 MB ...
136 modules
2024-08-25 01:44:33,314 - ControlNet - INFO - ControlNet Method lineart_standard (from white bg & black line) patched.
To load target model KModel
Begin to load 1 model
[Unload] Trying to free 17148.88 MB for cuda:0 with 0 models keep loaded ...
[Unload] Current free memory is 7781.50 MB ...
[Unload] Unload model JointTextEncoder
[Memory Management] Current Free GPU Memory: 9540.69 MB
[...]

Issue #1392, opened 2024-08-22
controllite fail on sdxl


laiyanyi:
```
Traceback (most recent call last):
  File "D:\SD\webui_forge_cu121_torch231\webui\modules_forge\main_thread.py", line 30, in work
    self.result = self.func(*self.args, **self.kwargs)
  File "D:\SD\webui_forge_cu121_torch231\webui\modules\txt2img.py", line 112, in txt2img_function
    processed = processing.process_images(p)
  File "D:\SD\webui_forge_cu121_torch231\webui\modules\processing.py", line 815, in process_images
    res = process_images_inner(p)
  File "D:\SD\webui_forge_cu121_torch231\webui\modules\processing.py", line 958, in process_images_inner
    samples_ddim = p.sample(conditioning=p.c, unconditional_conditioning=p.uc, seeds=p.seeds, subseeds=p.subseeds, subseed_strength=p.subseed_strength, prompts=p.prompts)
  File "D:\SD\webui_forge_cu121_torch231\webui\modules\processing.py", line 1329, in sample
    samples = self.sampler.sample(self, x, conditioning, unconditional_conditioning, image_conditioning=self.txt2img_image_conditioning(x))
  File "D:\SD\webui_forge_cu121_torch231\webui\modules\sd_samplers_kdiffusion.py", line 238, in sample
    samples = self.launch_sampling(steps, lambda: self.func(self.model_wrap_cfg, x, extra_args=self.sampler_extra_args, disable=False, callback=self.callback_state, **extra_params_kwargs))
  File "D:\SD\webui_forge_cu121_torch231\webui\modules\sd_samplers_common.py", line 272, in launch_sampling
    return func()
  File "D:\SD\webui_forge_cu121_torch231\webui\modules\sd_samplers_kdiffusion.py", line 238, in <lambda>
    samples = self.launch_sampling(steps, lambda: self.func(self.model_wrap_cfg, x, extra_args=self.sampler_extra_args, disable=False, callback=self.callback_state, **extra_params_kwargs))
  File "D:\SD\webui_forge_cu121_torch231\system\python\lib\site-packages\torch\utils\_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "D:\SD\webui_forge_cu121_torch231\webui\k_diffusion\sampling.py", line 595, in sample_dpmpp_2m
    denoised = model(x, sigmas[i] * s_in, **extra_args)
[...]

PR #1665, opened 2024-09-02
Set controlllite module dtype to model dtype

wcole3:
Sets the dtype of the controlllite controlnet to match the dtype of the model; addresses issue reported in #1641 #1488 #1392 for builtin kohya controlllite models and the [bdsqlsz](https://huggingface.co/bdsqlsz/qinglong_controlnet-lllite) models as far as I can tell (hopefully other controlllite's as well).

Full disclosure, I'm not sure what the quality consequences of changing the dtype, only that this change causes the error to go away.  Would appreciate @DenOfEquity's eyes on this since they reported not being able to reproduce the issue.

 No cnet (Base Model: protovisionXL)
 
image

Control Image

image

Depth Kohya_controlllite

image
image

Depth bdsqlsz

image

Canny bdsqlsz

image
image

### DIFF
--- a/extensions-builtin/sd_forge_controlllite/lib_controllllite/lib_controllllite.py
+++ b/extensions-builtin/sd_forge_controlllite/lib_controllllite/lib_controllllite.py
@@ -214,8 +214,6 @@ class LLLiteModule(torch.nn.Module):
 
         if self.cond_emb is None:
             # print(f"cond_emb is None, {self.name}")
-            # test bad idea
-            #self.cond_image = self.cond_image.view(dtype=x.dtype)
             cx = self.conditioning1(self.cond_image.to(x.device, dtype=x.dtype))
             if not self.is_conv2d:
                 # reshape / b,c,h,w -> b,h*w,c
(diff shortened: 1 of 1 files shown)


DenOfEquity:
These changes make it fail for me, same error but order reversed: `RuntimeError: Input type (float) and bias type (struct c10::Half) should be the same`. I don't think storage_dtype is a safe attribute to use - it is not necessarily the processing dtype.

wcole3:
Glad I asked, I'll take another look

DenOfEquity:
Try setting the dtype of the LLLiteModule, line 66, from the dtype of the cond_image.

wcole3:
Sadly, that returns the original error.  Do you know if the UnetPatcher.model.diffusion_model.computation_dtype would be what we're looking for?  That seems to work for me, but I'm having trouble figuring out how to get my IDE to run with fp32

DenOfEquity:
You're a few minutes ahead of me, just spotted that by printing out the model vars - seems like it should be correct, it is different to storage_dtype for me. Commandline argument `--all-in-fp32`.

wcole3:
Cool, that did the trick, thanks!  Using `computation_dtype` it seems to be working in fp32 and fp16; I'll do some more testing before pushing up and reopening.

DenOfEquity:
Just being picky while I do a completely unnecessary double (triple+?) check: we can lose lines 217-218, the commented out test.

wcole3:
Yupyup, sorry, sloppy housekeeping.  I can remove the LLLiteModule dtype field too if you think it isn't needed, I mainly put it there for debugging

DenOfEquity:
line 139? Very Pythonic. Sure, then I'm happy to hit Merge. Thanks!

wcole3:
You got it, I removed the field.  I'll followup in the issues to see if it's resolved.  Cheers, thanks for your help.

PR #1827, opened 2024-09-15
Allow loading of .sft VAE files for Flux

bkosowski:
Allow loading of `.sft` VAE files that are commonly used with Flux models


suphamster:
Maybe it's time to add it to main branch coz 1 month to review several lines of code is a bit too much for my liking.

Issue #1535, opened 2024-08-27
[suggestion] add an iteration value to hiresfix


andy8992:
It's my favorite feature in the efficiency nodes for comfy. Being able to say, use these settings to run the image through hiresfix X amount of times using the applied settings.

So it can perform it 2 or 3 times at 1.25 scale or what have you. 

I can move this to another area for suggestions if there is one.

CyborgNinjaJesus:
I would also like this to be implemented if it's possible. 

The very popular realistic image made with Flux (the woman with the google badge doing a live presentation) was generated using this method in Comfy. There's a lora in civitai that claims that image was made using it but it appears to be a hoax, the author of the image posted his workflow and it involved multiple upscaling iterations as it's suggested here.

Haoming02:
Can't you just set the **Upscale by** to `1.25` in the **Hires. fix** settings, then click the `Upscale` button 3 times?

This way, you can also interrupt early if like the first or second iteraiton produces something unwanted.

andy8992:
Not that I have seen the resolution stays the same. Perhaps it does work this way if you queue them but I know it doesn't if you render, wait, hit the button again

andy8992:
> Can't you just set the **Upscale by** to `1.25` in the **Hires. fix** settings, then click the `Upscale` button 3 times?
> 
> This way, you can also interrupt early if like the first or second iteraiton produces something unwanted.

confirmed no, it just redoes the same original image

it makes logical sense that the scale factor applies to the current image selected but it is relative to the original resolution so you must drag the slider over and over and over again to go bigger

ideally, the slider would be relative to the current image in the render window OR there could be an iteration value you can choose ahead of time

would save me a ton of clicking

Haoming02:
Huh, so clicking the `Upscale` button again still only processes the original `txt2img` result?

TIL

andy8992:
> Huh, so clicking the `Upscale` button again still only processes the original `txt2img` result?
> 
> TIL

Yea, rather I think this is what it does

512x512 goes to 768x768 for example then you hit it again and it rerenders the 768 image again, but it doesn't apply the upscale. So it's using the new image but it isn't getting larger. You can tell because the colors of the image get progressively worse each time, at least they do for me and my settings. If it were using the original image I'd expect the colors to be the same each time.

Haoming02:
Oh yeah. I just tried it as well.

[...]

PR #1838, opened 2024-09-16
autoset width and height

DenOfEquity:
* When loading into img2img, added an option to automatically set Width and Height from image (Settings -> img2img ->After loading into Img2img, automatically update Width and Height).
* fixed **Resize by** *Scale* display, now correctly rounded to multiple of 8
* That display now updates on change, instead of on release, so it updates as the slider is dragged and when an image is loaded (if option enabled)
* Also fixed new width/height calculation inside img2img function, same fix, (affects infotext)


Alexdnk:
Broken? repeated use of hrfix quickbutton as discussed in [#1535](https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/1535)

Issue #1420, opened 2024-08-22
ControlNet AttributeError: 'KModel' object has no attribute 'noise_augmentor'


Melyns:
I'm trying to use Revision mode in Forge ControlNet extension, but i keep getting this error `AttributeError: 'KModel' object has no attribute 'noise_augmentor'`. 

The model `CLIP-ViT-bigG.safetensors` is located at `D:\AI\Forge\webui\models\ControlNetPreprocessor` folder (in case this is important information for diagnosis). I verified the hash of the model downloaded on my PC & the one stored at HuggingFace & they match. My only guess is an error in the forge code somewhere? I'd highly appreciate any assistance.

version: [f2.0.1v1.10.1-previous-409-gf7ab23b1](https://github.com/lllyasviel/stable-diffusion-webui-forge/commit/f7ab23b1cb5d2e17d33ecc8febe9313bf824d897)
python: 3.10.6
torch: 2.4.0+cu124
xformers: N/A

Here's my console log:

```
ControlNet - INFO - ControlNet Method CLIP-G (Revision) patched.
To load target model KModel
Begin to load 1 model
[Unload] Trying to free 7676.88 MB for cuda:0 with 0 models keep loaded ...
[Unload] Current free memory is 5682.18 MB ...
[Unload] Unload model CLIPVisionModelWithProjection
[Unload] Current free memory is 9201.43 MB ...
[Memory Management] Current Free GPU Memory: 9201.43 MB
[Memory Management] Required Model Memory: 4897.05 MB
[Memory Management] Required Inference Memory: 1024.00 MB
[Memory Management] Estimated Remaining GPU Memory: 3280.38 MB
Moving model(s) has taken 2.88 seconds
  0%|                                                                                           | 0/20 [00:00<?, ?it/s]
Traceback (most recent call last):
  File "D:\AI\Forge\webui\modules_forge\main_thread.py", line 30, in work
    self.result = self.func(*self.args, **self.kwargs)
  File "D:\AI\Forge\webui\modules\txt2img.py", line 112, in txt2img_function
    processed = processing.process_images(p)
  File "D:\AI\Forge\webui\modules\processing.py", line 815, in process_images
    res = process_images_inner(p)
  File "D:\AI\Forge\webui\modules\processing.py", line 958, in process_images_inner
[...]

PR #1860, opened 2024-09-18
temp fix for preprocessor_revision.py

Melyns:
This is a temporary fix for the bug with controlnet revision `ControlNet AttributeError: 'KModel' object has no attribute 'noise_augmentor'`. https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/1420

I added a function that checks if the noise_augmentor attribute is present. If it is not, it prints a warning message and skips the noise augmentation step. Revision mode can still proceed even if the noise_augmentor attribute error occurs. I hid the `Noise Augmentation` slider since this fix disables that step.

### DIFF
--- a/extensions-builtin/forge_preprocessor_revision/scripts/preprocessor_revision.py
+++ b/extensions-builtin/forge_preprocessor_revision/scripts/preprocessor_revision.py
@@ -7,7 +7,14 @@ from modules_forge.shared import add_supported_preprocessor
 
 def revision_conditioning_modifier(model, x, timestep, uncond, cond, cond_scale, model_options, seed):
     revision_conditions = model_options['revision_conditions']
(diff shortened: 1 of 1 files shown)


DenOfEquity:
Good work.
I'd not bother with the warning/error message - any user that reads it can't do anything about it anyway, and it messes up the progress display. Then the initial check could be replaced with `noise_augmentor = getattr(model, 'noise_augmentor', None)`.

Melyns:
> Good work. I'd not bother with the warning/error message - any user that reads it can't do anything about it anyway, and it messes up the progress display. Then the initial check could be replaced with `noise_augmentor = getattr(model, 'noise_augmentor', None)`.

thank you, great suggestion. i applied what you have recommended, its much better now.

PR #1892, opened 2024-09-22
partial revert to fix compatible issue

wkpark:
* fix `submit()`, `submit_*()`, `create_submit_args()`
* ~~add `console.warn = function(){};` to ignore bunch of warnings~~ extracted to another PR.

these `submit*()` are used several extensions by means of hook or direct call.
this fix revert some part of gradio 4 fix to make it more compatible with A1111.
for example, uddetailer use `submit.apply()` call directly.

### DIFF
--- a/javascript/ui.js
+++ b/javascript/ui.js
@@ -113,6 +113,17 @@ function get_img2img_tab_index() {
 function create_submit_args(args) {
     var res = Array.from(args);
 
+    // As it is currently, txt2img and img2img send back the previous output args (txt2img_gallery, generation_info, html_info) whenever you generate a new image.
+    // This can lead to uploading a huge gallery of previously generated images, which leads to an unnecessary delay between submitting and beginning to generate.
+    // I don't know why gradio is sending outputs along with inputs, but we can prevent sending the image gallery here, which seems to be an issue for some.
+
+    // If gradio at some point stops sending outputs, this may break something
+    if (Array.isArray(res[res.length - 4])) {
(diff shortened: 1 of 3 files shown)


javascript/gradio.js
```diff
@@ -5,3 +5,6 @@ Object.defineProperty(Array.prototype, 'toLowerCase', {
         return this;
     }
 });
+
+// disable console.warn
+console.warn = function(){};
```
chocolateboy, line 10: It would be better to fix the warnings rather than hiding them.

chocolateboy (changes requested):

javascript/gradio.js
wkpark: simply removed. `console.warn` is not related with this issue.

lllyasviel:
hi do you have some code links to what might be using this to help understanding

DenOfEquity:
Seems OK with txt2img and img2img, but broken with Extras - because Extras has three outputs instead of four.
Rather than having two versions of `create_submit_args`, or further modifying it to handle this case, IMO it would now be better to inline the fixes into `submit`, `submit_img2img`, and `submit_extras`; leaving the original function as is.

wkpark:
as you know, originally, `create_submit_args()` simply nullify gallery output array to fix strage gradio behavior.
it works ok with the A1111 but with the webui-forge, the output garbage array make some trouble.
so I simply drop the output array as forge originally intended and fix compatible issue for some extensions.

PR #1899, opened 2024-09-23
workaround fix to suppress console.warn

wkpark:
without this fix you may get too much noisy warnings in your chrome debug console. it's not forge's fault. its a gradio4 bug.

image

### DIFF
--- a/javascript/gradio.js
+++ b/javascript/gradio.js
@@ -5,3 +5,7 @@ Object.defineProperty(Array.prototype, 'toLowerCase', {
         return this;
     }
 });
+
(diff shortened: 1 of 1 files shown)


nor1c:
CPU usage still high

DenOfEquity:
IMO, this isn't something the webUI should be doing; even though I agree the warnings can be annoying.
I've found that the warning `Attempted to select a non-interactive or hidden tab.` can be avoided by giving gradio tabs and tabitems an `id`; but haven't found any answer for the much more common `Too many arguments provided for the endpoint.`

PR #1906, opened 2024-09-24
Allow split T5 & CLIP prompts for flux & add a separate T5 token counter

wbclark:
(full description coming shortly)

### DIFF
--- a/backend/diffusion_engine/base.py
+++ b/backend/diffusion_engine/base.py
@@ -51,7 +51,12 @@ class ForgeDiffusionEngine:
         pass
 
     def get_prompt_lengths_on_ui(self, prompt):
-        return 0, 75
+        return {
+            't5': 0,
+            't5_max': 255,
+            'clip': 0,
+            'clip_max': 75,
+        }
 
     def is_webui_legacy_model(self):
         return self.is_sd1 or self.is_sd2 or self.is_sdxl or self.is_sd3
--- a/backend/diffusion_engine/flux.py
+++ b/backend/diffusion_engine/flux.py
@@ -74,9 +74,22 @@ class Flux(ForgeDiffusionEngine):
 
     @torch.inference_mode()
     def get_learned_conditioning(self, prompt: list[str]):
+        prompt_t5 = []
+        prompt_l = []
+
+        for p in prompt:
+            if 'SPLIT' in p:
+                before_split, after_split = p.split('SPLIT', 1)
+                prompt_t5.append(before_split.strip())
+                prompt_l.append(after_split.strip())
+            else:
+                prompt_t5.append(p)
+                prompt_l.append(p)
+
         memory_management.load_model_gpu(self.forge_objects.clip.patcher)
-        cond_l, pooled_l = self.text_processing_engine_l(prompt)
-        cond_t5 = self.text_processing_engine_t5(prompt)
+
+        cond_l, pooled_l = self.text_processing_engine_l(prompt_l)
+        cond_t5 = self.text_processing_engine_t5(prompt_t5)
         cond = dict(crossattn=cond_t5, vector=pooled_l)
 
         if self.use_distilled_cfg_scale:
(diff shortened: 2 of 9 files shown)


anae-git:
https://github.com/DenOfEquity/forgeFlux_dualPrompt

wbclark:
Hey @DenOfEquity! I noticed your forgeFlux_dualPrompt extension and saw that we're working on similar ideas around multi-prompt support. I recently submitted this PR for dual prompt support and a T5 token counter for Flux models, and I'm also thinking about future improvements such as providing a complete tokenization breakdown for the user. That would be beneficial for users of single encoder models too, although it's an interesting challenge to do that in a way that's not disruptive to the existing interface...

Anyway, I'd love to hear your thoughts on the approach in this PR and if you're interested in collaborating. Before I started working on this, I looked through past issues and PRs here but didn’t see anything similar—so I'm curious now if this was something you and @lllyasviel had previously discussed, especially in terms of native support vs. an extension?

DenOfEquity:
Hi. I haven't had time to test this yet, and I'd also like to see if lllyasviel has new opinions on it. IMO this keyword method is ideal as there's no major UI changes.
There's some brief discussion https://github.com/lllyasviel/stable-diffusion-webui-forge/discussions/1182, where lllyasviel wasn't sure if direct inclusion was worth it and my brief testing showed some potential but not definitive value. I think the best comparison is between `{prompt1} {prompt2}` and `{prompt1} SPLIT {prompt2}`.

Method should be extensible to at least sdxl (clip-l, clip-g).
Does it work with the prompt-bracket-checker builtin extension?

wbclark:
Thanks for the initial feedback.

> IMO this keyword method is ideal as there's no major UI changes.

I agree, and there is no API change either. It all flows through the same prompt argument as before, and there is no change if the SPLIT keyword is not provided. Plus, users are already familiar with this paradigm (AND, BREAK, etc)... but mostly, I implemented it this way so that the change would be small and not increase maintenance burden.

> There's some brief discussion https://github.com/lllyasviel/stable-diffusion-webui-forge/discussions/1182, where lllyasviel wasn't sure if direct inclusion was worth it and my brief testing showed some potential but not definitive value. I think the best comparison is between {prompt1} {prompt2} and {prompt1} SPLIT {prompt2}.

One reason I included the tokenization breakdown in my example is that I think it makes it much easier to see why we should care about the ability to specify different prompts per encoder.

I've been working with print debugging to see the tokenization in the terminal, but I want to find the right way to show this to the user, because understanding this will help them to write better prompts (even when using single encoder models).

Two designs I'm thinking about -- 1. show prompt tokenization on mouseover of the token counter, or 2. add a new accordion (collapsed by default to minimize visual disruption) which shows the tokenization once expanded.

I think another good comparison to try is simply `prompt` vs. `prompt SPLIT` vs. `SPLIT prompt`, which helps to understand how truly different the usage of each prompt is in the Flux model architecture.

Anyway, since this project aims to be more research focused, I think there is also a good argument for exposing the functionality to users, as they may discover some methods and applications that we did not consider.

> Method should be extensible to at least sdxl (clip-l, clip-g).

I haven't experimented with that yet, and while my gut intuition is that it might be less valuable than the Flux case due to architectural differences, it would still be a _very_ small change to implement as you are pointing out.

Thinking about it also has me rethinking a few things, like naming the token counters primary/secondary instead of CLIP/T5, and adding the additional counter to the negative prompt as well.

That might actually simplify things even further -- it may be possible to even treat both counters as a single UI element, and do most all changes `backend/` and `javascript/`, to further limit the necessary change in `modules/`. I will investigate this approach soon.

Plus some of my tests with setting CFG != 1 on Flux have me thinking perhaps negative prompts aren't totally DoA there after all.

>  Does it work with the prompt-bracket-checker builtin extension?

No, it requires a small adjustment... and I should also test how embeddings are handled just to make sure. Thanks for pointing this out.

PR #1952, opened 2024-09-30
Fix for use original name with batch img2img

dalapenko:
Hello. My friends use this tools and asked me to look why when use batch in img2img original name can't be save.

I'm found a bug when in **images.py** don't apply _samples_filename_pattern_ from img2img.py because pattern was write into **override_settings** but read only from **options** (**opts**). Fixed it.

For upload tab original filename never saved because **output_dir** forced set to empty. It's may be justified because we will override already existing files in some cases. I'm added new checkbox to upload tab for use original filename if it's really necessary.

### DIFF
--- a/modules/images.py
+++ b/modules/images.py
@@ -674,9 +674,9 @@ def save_image(image, path, basename, seed=None, prompt=None, extension='png', i
         if short_filename or seed is None:
             file_decoration = ""
         elif opts.save_to_dirs:
-            file_decoration = opts.samples_filename_pattern or "[seed]"
+            file_decoration = opts.samples_filename_pattern or p.override_settings.get("samples_filename_pattern") or "[seed]"
         else:
-            file_decoration = opts.samples_filename_pattern or "[seed]-[prompt_spaces]"
+            file_decoration = opts.samples_filename_pattern or p.override_settings.get("samples_filename_pattern") or "[seed]-[prompt_spaces]"
 
         file_decoration = namegen.apply(file_decoration) + suffix
 
--- a/modules/img2img.py
+++ b/modules/img2img.py
@@ -122,16 +122,14 @@ def process_batch(p, input, output_dir, inpaint_mask_dir, args, to_scale=False, 
         if output_dir:
             p.outpath_samples = output_dir
             p.override_settings['save_to_dirs'] = False
-            p.override_settings['save_images_replace_action'] = "Add number suffix"
-            if p.n_iter > 1 or p.batch_size > 1:
-                p.override_settings['samples_filename_pattern'] = f'{image_path.stem}-[generation_number]'
-            else:
-                p.override_settings['samples_filename_pattern'] = f'{image_path.stem}'
+
+        if opts.img2img_batch_use_original_name:
+            filename_pattern = f'{image_path.stem}-[generation_number]' if p.n_iter > 1 or p.batch_size > 1 else f'{image_path.stem}'
+            p.override_settings['samples_filename_pattern'] = filename_pattern
 
         proc = modules.scripts.scripts_img2img.run(p, *args)
 
         if proc is None:
-            p.override_settings.pop('save_images_replace_action', None)
             proc = process_images(p)
 
         if not discard_further_results and proc:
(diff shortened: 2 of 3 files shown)


DenOfEquity:
Hi, thanks for your PR.
I've verified that the override handling is covered by #1465.
I'm not sure about changing the behaviour of the *Upload* tab; risk of unintended overwrites seems relatively high compared to the *From directory* tab, where the user has to manually set the save directory. Opinions welcome.

dalapenko:
@DenOfEquity Yes, the risk of unintentional overwriting can be high and that's why I made the default checkbox off with a warning about the risks. It seemed to me that the current behavior is very unclear to the user. It's hard to guess without looking at the code that the original name won't be saved on the upload tab, but on the tab from a dir it will be. Perhaps it's worth adding at least just an informational text

DenOfEquity:
I've done some more poking around. The original code has no risk of overwrite, because it sets another override `p.override_settings['save_images_replace_action'] = "Add number suffix"`.

So the problem in this area is, as you say, that saving behaviour is inconsistent and unpredictable to anyone without prior knowledge. I think the solution is to add an option in Settings to use original filename that can be applied to both *Upload* and *From directory* tabs.

shared_options.py, insert line 233 (end of the img2img block):
`    "img2img_batch_use_original_name": OptionInfo(False, "Save using original filename in img2img batch. Applies to 'Upload' and 'From directory' tabs."),`

img2img.py, insert line 125 (before the override line mentioned earlier):
`        if opts.img2img_batch_use_original_name:`

It's a moot point at the moment anyway, with override settings not being used.

dalapenko:
Tried to do as you say:

* pulled a fresh master from your repo
* added the img2img_batch_use_original_name setting to settings/img2img

but I do not understand about

> img2img.py, add line 125 (before the override line mentioned above):
> if opts.img2img_batch_use_original_name:

What do you mean? main problem as i see - some logic require to have output dir whenever its empty for upload tab.

In this round I'm separating logic with output dir setting from save_images_replace_action override (it's option not working as I see) and setting samples_filename_pattern for saving file.

If option img2img_batch_use_original_name is true, we do not add any data to filename programatically for each tab.

DenOfEquity:
The current/old logic uses `output_dir` to determine if original filenames should be reused. That is why it only works on the 'From directory' tab.
My suggested new logic uses the new setting to determine if the original filename is used: not dependent on `output_dir`; consistent behaviour on both batch tabs; and lets the user decide if they want to use original name(s) or not. *Settings > Saving images/grids > Saving the image to an existing file* then controls whether overwrites are allowed.
You are correct that the override 'save_images_replace_action' doesn't affect things as I assumed in my previous comment, it seems in normal batch operation that the override gets removed later. I'm not sure what's going on there. But I think this is fine - the setting mentioned earlier allows the user to control overwrites.
In my testing, this all works logically and consistently. The option description could be updated to show that overwrites are possible: `    "img2img_batch_use_original_name": OptionInfo(False, "Save using original filename in img2img batch. Applies to 'Upload' and 'From directory' tabs.").info("Warning: overwriting is possible, based on Settings > Saving images/grids > Saving the image to an existing file."),`
As I read it, your version always uses original filename (no choice, but consistent), and the option does nothing because that override doesn't take effect.

dalapenko:
@DenOfEquity 

- Add warning info to option in settings
- Remove forced _save_images_replace_action_ from img2img logic (now it depends on settings like in option warning)
- Returned fix with check_p.override_settings.get("samples_filename_pattern")_ from my previous iteration. Fix from https://github.com/lllyasviel/stable-diffusion-webui-forge/pull/2027 not work when choose any scripts for processing. Name when save still not use apply _samples_filename_pattern_.

now it's work consistent between tabs and depends on settings option

DenOfEquity:
Thanks for your work on this. Point 3 is a good catch.
Line 133 in your updated img2img.py is now unnecessary? As the override it removes is never set.

dalapenko:
@DenOfEquity yes, you're right - remove it. thx

rushuna86:
when running Adetailer through i2i with this commit, unable to save the result of Adetailer that's outputted. Checkout to 05b01da and it saves perfectly fine.

wensleyoliv:
When using X/Y/Z plot to change checkpoints with Checkpoint name, i'm getting the following error.
```
Traceback (most recent call last):
  File "D:\webui-forge\webui\modules_forge\main_thread.py", line 30, in work
    self.result = self.func(*self.args, **self.kwargs)
  File "D:\webui-forge\webui\modules\txt2img.py", line 121, in txt2img_function
    processed = modules.scripts.scripts_txt2img.run(p, *p.script_args)
  File "D:\webui-forge\webui\modules\scripts.py", line 792, in run
    processed = script.run(p, *script_args)
  File "D:\webui-forge\webui\scripts\xyz_grid.py", line 832, in run
    images.save_image(processed.images[g], p.outpath_grids, "xyz_grid", info=processed.infotexts[g], extension=opts.grid_format, prompt=processed.all_prompts[adj_g], seed=processed.all_seeds[adj_g], grid=True, p=processed)
  File "D:\webui-forge\webui\modules\images.py", line 677, in save_image
    file_decoration = opts.samples_filename_pattern or p.override_settings.get("samples_filename_pattern") or "[seed]"
AttributeError: 'Processed' object has no attribute 'override_settings'
'Processed' object has no attribute 'override_settings'
```

dalapenko:
create PR with possible fix https://github.com/lllyasviel/stable-diffusion-webui-forge/pull/2253

thx for fast feedback

Issue #1940, opened 2024-09-29
/sdapi/v1/txt2img override_settings parameter don't override checkpoint


wangyafeng1990:

vaskers5:
+

vaskers5:
+ - the same

altoiddealer:
Frankly, I don't think `override_settings` is working at all right now.

As a workaround, you can POST your `override_settings` dict to the `/sdapi/v1/options` endpoint and it will change the image model and settings.

vaskers5:
Open pr with fix for this issue

altoiddealer:
> Open pr with fix for this issue

I do what I can :P

vaskers5:
> > Open pr with fix for this issue
> 
> I do what I can :P

I mean, I've opened a pr with a fix for this problem and I'm waiting for a review)

altoiddealer:
> > > Open pr with fix for this issue
> > 
> > 
> > I do what I can :P
> 
> I mean, I've opened a pr with a fix for this problem and I'm waiting for a review)

Awesome!   Can’t wait to check it out

PR #1961, opened 2024-10-02
Bug fix override settings sdapi

vaskers5:


vaskers5:
Fixed this issue:

https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/1940

altoiddealer:
Ehh... I was excited about this PR, but I don't think this is the correct fix.

What this does is makes it so that an `override_settings` dict will modify the current settings - a persistent change.

Historically, the purpose of `override_settings` is to only temporarily modify the settings for the current request ("Override the current settings").

vaskers5:
got  u, I will try to fix it in the right way

altoiddealer:
@vaskers5 I'm sorry but I was feeling a bit ambitious tonight, and was able to solve this.

I also made an additional improvement regarding the handling of modules via `options` and `override_settings`

https://github.com/lllyasviel/stable-diffusion-webui-forge/pull/2027

DenOfEquity:
closed as replaced by #2027

PR #2027, opened 2024-10-11
API Improvements: Modules Change AND Restore override_settings

altoiddealer:
## Improve API Modules Change:

**Previously**, the complete file path was required for each module specified in the 'forge_additional_modules' list.

```
'forge_additional_modules' = [
"C:\\0_SD\\stable-diffusion-webui-forge\\models\\text_encoder\\ae.safetensors",
"C:\\0_SD\\stable-diffusion-webui-forge\\models\\text_encoder\\clip_l.safetensors"
]
```

**Now**, it only requires the module name (like `sd_model_checkpoint`)

```
'forge_additional_modules' = [
"ae.safetensors",
"clip_l.safetensors"
]
```

In the process, I added a positional argument to prevent spamming `refresh_model_loading_parameters()`

---

## Restore override_settings:

I found the corresponding code in A1111, and determined the correct lines to integrate `override_settings` handling back in.

Although the code is a bit verbose, this is customized for the Forge environment to gracefully handle memory, checkpoints, and modules.

### DIFF
--- a/modules/api/api.py
+++ b/modules/api/api.py
@@ -692,10 +692,13 @@ class Api:
         memory_keys = ['forge_inference_memory', 'forge_async_loading', 'forge_pin_shared_memory']
 
         for k, v in req.items():
+            # options for memory/modules are set in their dedicated functions
             if k in memory_keys:
-                memory_changes[k] = v
+                mem_key = k[len('forge_'):] # remove 'forge_' prefix
+                memory_changes[mem_key] = v
             elif k == 'forge_additional_modules':
                 main_entry.modules_change(v, refresh_params=False) # refresh_model_loading_parameters() --- applied in checkpoint_change()
+            # set all other options
             else:
                 shared.opts.set(k, v, is_api=True)
 
--- a/modules/api/api.py
+++ b/modules/api/api.py
@@ -692,9 +692,12 @@ class Api:
         memory_keys = ['forge_inference_memory', 'forge_async_loading', 'forge_pin_shared_memory']
 
         for k, v in req.items():
-            shared.opts.set(k, v, is_api=True)
             if k in memory_keys:
                 refresh_memory = True
+            if k == 'forge_additional_modules':
+                main_entry.modules_change(v, refresh_params=False)
+            else:
+                shared.opts.set(k, v, is_api=True)
 
         main_entry.checkpoint_change(checkpoint_name)
         # shared.opts.save(shared.config_filename) --- applied in checkpoint_change()
--- a/modules/api/api.py
+++ b/modules/api/api.py
@@ -688,23 +688,22 @@ class Api:
         if checkpoint_name is not None and checkpoint_name not in sd_models.checkpoint_aliases:
             raise RuntimeError(f"model {checkpoint_name!r} not found")
         
-        refresh_memory = False
+        memory_changes = {}
         memory_keys = ['forge_inference_memory', 'forge_async_loading', 'forge_pin_shared_memory']
 
         for k, v in req.items():
             if k in memory_keys:
-                refresh_memory = True
-            if k == 'forge_additional_modules':
-                main_entry.modules_change(v, refresh_params=False)
+                memory_changes[k] = v
+            elif k == 'forge_additional_modules':
+                main_entry.modules_change(v, refresh_params=False) # refresh_model_loading_parameters() --- applied in checkpoint_change()
             else:
                 shared.opts.set(k, v, is_api=True)
 
         main_entry.checkpoint_change(checkpoint_name)
         # shared.opts.save(shared.config_filename) --- applied in checkpoint_change()
 
-        if refresh_memory:
-            model_memory = main_entry.total_vram - shared.opts.forge_inference_memory
-            main_entry.refresh_memory_management_settings(model_memory, shared.opts.forge_async_loading, shared.opts.forge_pin_shared_memory)
+        if memory_changes:
+            main_entry.refresh_memory_management_settings(**memory_changes)
 
         return
 
--- a/modules/processing.py
+++ b/modules/processing.py
@@ -32,6 +32,7 @@ from einops import repeat, rearrange
 from blendmodes.blend import blendLayers, BlendType
 from modules.sd_models import apply_token_merging, forge_model_reload
 from modules_forge.utils import apply_circular_forge
+from modules_forge import main_entry
 from backend import memory_management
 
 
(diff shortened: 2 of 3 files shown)


altoiddealer:
I'm working on some additional changes to reduce the complexity of memory management - mainly, simplify the logic behind how the UI `GPU Weights` has an inverted effect compared to providing `forge_inference_memory` via the API.

Should have this updated shortly.

altoiddealer:
I updated the `refresh_memory_management_settings()` function to be more flexible:
- No longer requires any values to be provided.
- Will factor any provided values, otherwise factor current setting
- If a precalculated `model_memory` value is passed (via UI slider), does not perform redundant memory calculation.

Added a `ui_refresh_memory_management_settings()` function to convert the gradio ui component values list to key words.

Overall, the code for managing special settings (memory, modules, checkpoints) via `options` API endpoint and `override_settings` parameter is now much more clear and simplified, without having to dabble with changing the UI related code.

psydok:
Could you please advise, am I understanding correctly that now `override_settings` will change the service settings as a whole, not just for the duration of the request as it was in automatic1111? If I'm not mistaken.

altoiddealer:
No, this PR will make override_settings behave as it should (temporarily for current request only).

As part of this PR I also made it so that the list of modules only has to be their names, instead if full file paths (pertinent to both /options/ endpoint and override_settings)

DenOfEquity:
Seems good to me; tested, no issues found with normal webui use. I've held off merging in case lllyasviel still has concerns about reinstating overrides. But that was two months ago, it seems reasonable to go ahead now. And by 'now', I mean when I get online tomorrow rather than right now, when I need to go to sleep.

altoiddealer:
Oh wow, I completely forgot that you had a similar PR, and in limbo for 2 months.  Well, it does seem like most of the other loose strings have been tied up, most of which by you :) Now seems like a great time to restore this highly demanded feature

uorufu:
> Seems good to me; tested, no issues found with normal webui use. I've held off merging in case lllyasviel still has concerns about reinstating overrides. But that was two months ago, it seems reasonable to go ahead now. And by 'now', I mean when I get online tomorrow rather than right now, when I need to go to sleep.

As this is the only thing that I've still been using automatic1111 for I've been eagerly looking forward to this, however I've tried the new override patch out for adetailer's "Use separate checkpoint" and it seems to try and do something but still generates with the original model.

One face adetailer without separate checkpoint:
``` 
0: 640x640 1 face, 5.5ms
Speed: 3.0ms preprocess, 5.5ms inference, 1.0ms postprocess per image at shape (1, 3, 640, 640)
[Unload] Trying to free 1024.00 MB for cuda:0 with 0 models keep loaded ... Current free memory is 37415.43 MB ... Done.
Cleanup minimal inference memory.
SwinIR: 100%|################################################################################################################################################################################################| 9/9 [00:02<00:00,  3.69it/s]
[Unload] Trying to free 3534.00 MB for cuda:0 with 1 models keep loaded ... Current free memory is 37391.43 MB ... Done.
[Unload] Trying to free 1024.00 MB for cuda:0 with 1 models keep loaded ... Current free memory is 37415.06 MB ... Done.
[Unload] Trying to free 1024.00 MB for cuda:0 with 1 models keep loaded ... Current free memory is 37414.45 MB ... Done.
[Unload] Trying to free 1310.72 MB for cuda:0 with 1 models keep loaded ... Current free memory is 37413.58 MB ... Done.
100%|######################################################################################################################################################################################################| 23/23 [00:09<00:00,  2.52it/s]
[Unload] Trying to free 4356.00 MB for cuda:0 with 1 models keep loaded ... Current free memory is 37401.23 MB ... Done.
```

And with separate checkpoint:
```
0: 640x640 1 face, 6.0ms
Speed: 2.0ms preprocess, 6.0ms inference, 2.0ms postprocess per image at shape (1, 3, 640, 640)
Model selected: {'checkpoint_info': {'filename': 'C:\\AI\\_models\\checkpoints\\dreamshaperXL_alpha2Xl10.safetensors', 'hash': 'c8afe2ef'}, 'additional_modules': ['C:\\AI\\_models\\vae\\sdxlVAE_sdxlVAE.safetensors'], 'unet_storage_dtype': None}
Using online LoRAs in FP16: True
[Unload] Trying to free 1024.00 MB for cuda:0 with 0 models keep loaded ... Current free memory is 37405.36 MB ... Done.
Cleanup minimal inference memory.
SwinIR: 100%|################################################################################################################################################################################################| 9/9 [00:02<00:00,  3.68it/s]
[Unload] Trying to free 3534.00 MB for cuda:0 with 1 models keep loaded ... Current free memory is 37381.36 MB ... Done.
[Unload] Trying to free 1024.00 MB for cuda:0 with 1 models keep loaded ... Current free memory is 37404.98 MB ... Done.
[Unload] Trying to free 1024.00 MB for cuda:0 with 1 models keep loaded ... Current free memory is 37404.37 MB ... Done.
[Unload] Trying to free 1310.72 MB for cuda:0 with 1 models keep loaded ... Current free memory is 37403.51 MB ... Done.
100%|######################################################################################################################################################################################################| 23/23 [00:09<00:00,  2.49it/s]
[Unload] Trying to free 4356.00 MB for cuda:0 with 1 models keep loaded ... Current free memory is 37402.55 MB ... Done.
Model selected: {'checkpoint_info': {'filename': 'C:\\AI\\_models\\checkpoints\\mix.safetensors', 'hash': '84ff2fd1'}, 'additional_modules': ['C:\\AI\\_models\\vae\\sdxlVAE_sdxlVAE.safetensors'], 'unet_storage_dtype': None}
Using online LoRAs in FP16: True
```
According to console it clearly louds the seperate checkpoint then makes the image and then changes back to original checkpoint... Yet the generated image is exactly the same and uses my mix.safetensors main checkpoint instead. In the above, both pictures are exactly the same output. DreamshaperXL_alpha2Xl10.safetensors does not appear to have been used.

altoiddealer:
@uorufu Ill try to reproduce the issue tonight - in the meantime could you share your payload?

Are you using alwayson_scripts for the adetailer params?  Could be something in there not being applied….

uorufu:
@altoiddealer 
Awesome, thanks!

altoiddealer:
> @altoiddealer Awesome, thanks!

I figured it out... big oversight on my part.  I falsely assumed that the model was actually changing due to what was being printed in CMD window.  I did not explicitly check the infotext in output results to confirm model change.

I **did** check resulting infotext for other override_settings: clip skip, forge_inference_memory, which I could see were being applied correctly.


Now in regards to the models - in the same code block for handling `override_settings` I see there are a few other functions dealing with models.  By moving these to occur **after** `override_settings` is applied, this is indeed making the model load correctly.  In the UI, when I generate another image it is correctly using the original model.

I'm going to do a bit more testing to be certain this is correct with no side effects.  It's looking good so far.

dimonnwc3:
@altoiddealer thanks for a very useful feature.
Since the param name is `sd_model_checkpoint` would it work for flux models as well?

ryankashi:
Oh wow, perfect timing for these updates! I was literally just messing around with the override_settings and specifically sd_model_checkpoint not actually switching.

Once the commit is in main I can do a test with flux as well for you if you'd like

altoiddealer:
@dimonnwc3 @ryankashi 

Until the new PR is merged you could just back up modules/processing.py and replace it with the changed file.

Changes to/from Flux all the same, just need to be sure to pass forge_additional_modules accordingly.

You can already change models via API using the options endpoint - just send the same dictionary you are using for override_settings.  This override_settings thing just makes it temporary while the options endpoint makes a persistent change

ryankashi:
@altoiddealer Just tested, I am also able to switch to a Flux model via the override_settings as well.

One small detail that I noticed different between sd-webui-forge and baseline Automatic1111 is that when you send back-to-back requests with the same ```sd_model_checkpoint```, it will try to load and unload the model every time, wheras baseline will keep it loaded.

This means that, if you are using the API to switch between models, repeated requests to the same model will be significantly faster with Automatic1111 vs sd-webui-forge!

altoiddealer:
@ryankashi I understand - the handling could be improved by only setting the model if it is different from most recent model.

I'll see how I can best implement this

altoiddealer:
@ryankashi Thanks for your comment - in the process of making a solution for this, I also cleaned up a lot of messy code.

A number of memory, module, and model loading functions are now only called when necessary and inbound setting is actually different than current.

https://github.com/lllyasviel/stable-diffusion-webui-forge/pull/2078

dhp666:
> 我正在进行一些额外的改变来降低内存管理的复杂性 - 主要是简化 UI与通过 API`GPU Weights`提供相比产生反转效果背后的逻辑。`forge_inference_memory`
> 
> 应该很快就会更新。

大佬，知道如何用api调用forge的flux模型吗

PR #2038, opened 2024-10-12
Unload all models control

DenOfEquity:
make sd_models.unload_model_weights() do what it claims (previously **pass**)
called in two places: API unloadapi(); and the Settings > Actions page.

### DIFF
--- a/modules/sd_models.py
+++ b/modules/sd_models.py
@@ -449,6 +449,7 @@ def reload_model_weights(sd_model=None, info=None, forced_reload=False):
 
 
 def unload_model_weights(sd_model=None, info=None):
+    memory_management.unload_all_models()
     pass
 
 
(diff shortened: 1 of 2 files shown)


altoiddealer:
I may be wrong about this, but I don't think this button / function actually does what it says it does.

- If I choose some new model in the UI and hit generate, it takes 1-2 seconds to unload the current model, and then ~15 seconds to load the new one.
- Subsequent generations occur almost instantly when pressing Generate.
- IF I press the "Unload All Models" button, then click Generate - again, generation occurs almost instantly.  I believe the most recent model must still reside in memory?

Arnaud3013:
If people are using extension "T5 on other device" like CPU, when changing models it does have lot of memory leak (vram on RAM, not GPU VRAM), it should be good to check if those people are using this extension

Issue #2051, opened 2024-10-13
preprocessor images not being returned in api response


dermesut:
title says it basically.
in the forge-version prior to this "new", constantly updated one controlnet images that were generated, be it canny, lineart, depth etc., were sent together with the resulting image and could be fetched in the response.
i don't know, why that's not the case anymore - i found it very useful and would love to have those back.

altoiddealer:
@dermesut To clarify:

- Is this for txt2img only?  Or, expected for img2img but not working?
- I see that img2img endpoint has the parameter `include_init_images` - Are you setting this to `True` for img2img request?
- I do not see `include_init_images` as a valid request param in any other WebUIs **for txt2img** (A1111 / ReForge / Forge old / etc)
- In any case, it shouldn't be too painful to retain your own inputs and do whatever you want with them post-gen.

Please clarify your expectations

altoiddealer:
> title says it basically.

Also clarify, I think you mean:

preprocessor images not being **returned** with api response

dermesut:
> > title says it basically.
> 
> Also clarify, I think you mean:
> 
> preprocessor images not being **returned** with api response

oh yes, absolutely. thank you. changed the title.

dermesut:
> @dermesut To clarify:
> 
> * Is this for txt2img only?  Or, expected for img2img but not working?
> * I see that img2img endpoint has the parameter `include_init_images` - Are you setting this to `True` for img2img request?
> * I do not see `include_init_images` as a valid request param in any other WebUIs **for txt2img** (A1111 / ReForge / Forge old / etc)
> * In any case, it shouldn't be too painful to retain your own inputs and do whatever you want with them post-gen.
> 
> Please clarify your expectations

be it txt2img or img2img, the api-response used to deliever a list of images: the resulting redering together with the resulting precprocessor images (the ones that can be seen in the webui as small previews beneath the big rendered image). a depth map if one had a depth-preprocessor activated, or any other preprocessor result, really.
one can still see them in the webui, but via api-response they are not delievered anymore. in forge "1" they were, in forge "2" they aren't (not sure, how to name those... pre- and post-experimental state maybe?)

altoiddealer:
Ok so I believe the setting which controls this in particular is the controlnet parameter `save_detected_map`

Do you have this set to `true` for each controlnet?

dermesut:
[...]

PR #2066, opened 2024-10-15
Include Extra Images via API

altoiddealer:
Solves https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/2051

In all (?) versions of A1111, and in maybe (?) some previous versions of Forge and/or ReForge, any and all "Extra Images" would be appended to the image results.  The most common type of Extra Images is ControlNet preprocessor maps.

I think that when ControlNet was integrated, this feature was left out by accident or there were simply bigger fish to fry.

This PR will now add extra_images to the images list for the API.

Some more explanation on what's going on...

The ControlNet preprocessor maps are initially being added to `extra_images` as numpy arrays.  I couldn't ascertain how the UI processes them, I was only able to trace back the UI behavior to [here](https://github.com/lllyasviel/stable-diffusion-webui-forge/blob/main/modules/txt2img.py#L134) - and with print statements I see the `processed.images` is (typically) a one PIL image object, and the `processed.extra_images` are a list of numpy arrays.  I don't know how the UI deals with this, but it does.  I had tried simply adding `+ processed.extra_images` in the API version, but it does not accept the numpy arrays and yields an invalid infotext error.

I made a little function that will update the `extra_images` by converting any img data which is numpy arrays to PIL image objects using `Image.fromarray()` - then, only Image objects will be set.  If there are any other types of arrays that could be converted and applied, the function could simply be updated later if necessary.

### DIFF
--- a/modules/api/api.py
+++ b/modules/api/api.py
@@ -21,7 +21,7 @@ from modules import sd_samplers, deepbooru, images, scripts, ui, postprocessing,
 from modules.api import models
 from modules_forge import main_entry
 from modules.shared import opts
-from modules.processing import StableDiffusionProcessingTxt2Img, StableDiffusionProcessingImg2Img, process_images
+from modules.processing import StableDiffusionProcessingTxt2Img, StableDiffusionProcessingImg2Img, process_images, process_extra_images
 from modules.textual_inversion.textual_inversion import create_embedding
 from PIL import PngImagePlugin
 from modules.realesrgan_model import get_realesrgan_models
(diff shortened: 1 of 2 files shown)


altoiddealer:
### Before:

Before


### After:

After

DenOfEquity:
I think the Gradio gallery accepts various forms of input, including numpy arrays. In the hiresfix quickbutton function I had to add a check, because attempting to upscale a control image would lead to an error later due to missing infotext.

Anyway, thanks again for the fix.

PR #2078, opened 2024-10-15
Improve options management

altoiddealer:
I began working on this PR with the intention of simply consolidating redundant options handling code.

However, things began clicking together very well, and I'm proud to say this is a big improvement.

First, I noticed that `/sdapi/v1/options` GET had its own `get_config()` function, and that there also exists a `get_config()` function in the **sysinfo** module. I tested and both return 100% literally the same value, except the one in **sysinfo** seemed to be in a more suitable location, and also had more concise code.  I updated `/sdapi/v1/options` GET to simply calls that function.

From there, **sysinfo** module seemed like the best place to house a new, more flexible and robust `set_config()` function. This function achieves the following:
  - obsoletes redundant code from `/sdapi/v1/options` POST and `override_settings`
  - skips updating values that are unchanged
  - has flexible args for both API and UI, so it could possibly obsolete additional functions (but I didn't want ton get too carried away)

Additionally, `checkpoint_change()` and `modules_change()` now have positional args to dictate whether `refresh_model_loading_parameters()` and `shared.opts.save()` should be called.  These default to True (mainly, when called from UI).

- Updated `modules_change()` so that values in `forge_additional_modules` can be provided either as file path, or just the module name.

- Most importantly, `refresh_model_loading_parameters()` is now only called ONCE per request, and **only** if necessary.

- It is now much easier to call `shared.opts.save()` as needed

### DIFF
--- a/modules/api/api.py
+++ b/modules/api/api.py
@@ -19,7 +19,6 @@ from secrets import compare_digest
 import modules.shared as shared
 from modules import sd_samplers, deepbooru, images, scripts, ui, postprocessing, errors, restart, shared_items, script_callbacks, infotext_utils, sd_models, sd_schedulers
 from modules.api import models
-from modules_forge import main_entry
 from modules.shared import opts
 from modules.processing import StableDiffusionProcessingTxt2Img, StableDiffusionProcessingImg2Img, process_images, process_extra_images
 from modules.textual_inversion.textual_inversion import create_embedding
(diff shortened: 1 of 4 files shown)


DenOfEquity:
Thank you again, this is excellent IMO. Also helped improve and simplify my forthcoming restoration of highres fix checkpoints.

altoiddealer:
Thanks!  You beat me by 1 minute - there is one little flaw that needs to be corrected… I’ll be pushing another small PR soon.  If “forge_additional_modules” are provided but is unchanged, it can revert the flag to reload model params.

Issue #2082, opened 2024-10-16
It's now driving me mad... when changing the Diffusion in Low Bits, it often loads the wrong model, one i loaded much earlier


andy8992:
This has been a minor inconvenience but I'm testing a lot of settings right and now it's happened like 5 times within the hour.

I get the terminal message about a new model being selected despite me not selecting a new one. It keeps loading juggernaut despite a different model being selected in the drop down, it's just reverting to a model i loaded earlier.

To fix i have to swap models, then swap again.

DenOfEquity:
You should find this fixed, as of a few days ago.

Issue #2034, opened 2024-10-12
Checkpoint, clip skip and vae settings are not saved between sessions.


Symbiomatrix:
Could be a side effect of my fiddling, but none of the gr.state variables are getting saved to the current value.
When one presses "Apply settings" (ie during run_settings), the value which these three gr.state components sends is always none, thus upon reload they'll be set to the defaults - sd_model_checkpoint to null then the first in the list, CLIP_stop_at_last_layers to 1, sd_vae to Automatic.
As such, I think it would be preferable to exclude any gr.state components from the standard save.

If anyone with a more standard install can confirm this glitch, that'd be swell.

PR #2090, opened 2024-10-16
prevent (Managed by Forge) settings from reverting to default

DenOfEquity:
#2082
also thanks to #2034

Also made CLIP skip setting save to config when changed. Previously it was not saved on change, but would be saved when another saveable setting was changed (for example, the checkpoint). That felt unpredictable and inconsistent.

### DIFF
--- a/modules/ui_settings.py
+++ b/modules/ui_settings.py
@@ -83,6 +83,10 @@ class UiSettings:
             if comp == self.dummy_component:
                 continue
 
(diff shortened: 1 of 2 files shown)


Symbiomatrix:
Wouldn't it be preferable to prevent any state components being saved, like so? These three are just a symptom.
```
            if isinstance(comp, gr.State):
                continue
```

PR #2092, opened 2024-10-16
prevent undoing refresh model load params

altoiddealer:
In my last PR (https://github.com/lllyasviel/stable-diffusion-webui-forge/pull/2078), there was a flaw in the logic pertaining to `forge_additional_modules`, which could revert the flag `should_refresh_model_loading_params` back to `False`

Change ensures `should_refresh_model_loading_params()` is called when needed.

Improved code clarity.

### DIFF
--- a/modules/sysinfo.py
+++ b/modules/sysinfo.py
@@ -237,7 +237,9 @@ def set_config(req: dict[str, Any], is_api=False, run_callbacks=True, save_confi
             main_entry.checkpoint_change(v, save=False, refresh=False)
             should_refresh_model_loading_params = True
         elif k == 'forge_additional_modules':
-            should_refresh_model_loading_params = main_entry.modules_change(v, save=False, refresh=False)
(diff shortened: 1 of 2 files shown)


altoiddealer:
@DenOfEquity quick question / observation...

I noticed that the `refresh_model_loading_parameters()` is **Not** called when any of the options pertaining to memory management are modified (`'forge_inference_memory', 'forge_async_loading', 'forge_pin_shared_memory'`) via `refresh_memory_management_settings()`

I'm uncertain if this is an oversight or not?  Seems to me like the models/modules/memory settings all primary components in model loading behavior?

DenOfEquity:
I interpret those as inference settings, not loading settings. Calling it for those settings would just be redundant? Note that changing *Diffusion in Low Bits* does then call `refresh_model_loading_parameters`, because that is a loading setting.
But maybe I'm slow today; needed a second look at the PR diffs before I noticed why it's different :)

PR #2099, opened 2024-10-17
Clipskip override from infotext

DenOfEquity:
- added overrides from infotext for clip skip, also checkpoint override, also vae/te, also vae (old form).

### DIFF
--- a/modules/infotext_utils.py
+++ b/modules/infotext_utils.py
@@ -526,7 +526,7 @@ def get_override_settings(params, *, skip_fields=None):
         if setting_name == "sd_model_checkpoint" and shared.opts.disable_weights_auto_swap:
             continue
 
-        if setting_name in ["Module 1", "Module 2", "Module 3"] and shared.opts.disable_weights_auto_swap:
+        if setting_name == "forge_additional_modules" and shared.opts.disable_weights_auto_swap:
             continue
 
         v = shared.opts.cast_value(setting_name, v)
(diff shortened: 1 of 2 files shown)


altoiddealer:
I have not looked into this deeply yet, but I can see that the behavior of sending Modules from infotext is hard coded 1-3.

If there is some way to make it dynamic with **for loops** like

```
for k, v in infotext:
     if k.startswith(`Module`):
          options.append(v)
````

Or similar logic.  Because there really isn't a limit to how many modules can be used

**Comment by altoiddealer** [comment]
Haven't looked but I would also hope the infotext is written dynamically with something like

```
for index, module in enumerate(`forge_additional_modules`):
      infotext.append(f"Module {index}: {module}")
```

Or similar

**Comment by DenOfEquity** [comment]
Good point, I'd completely overlooked that. There's some other sloppiness too: don't need to include the modules in **shared_options**; better way of getting filename without extension.

**Comment by altoiddealer** [comment]
Looks good to me - I am suggesting a minor update for these lines starting at line 537:

```
        if setting_name == "sd_model_checkpoint" and shared.opts.disable_weights_auto_swap:
            continue

        if setting_name == "forge_additional_modules": 
            if shared.opts.disable_weights_auto_swap:
                continue
```

Suggested change:

```
        if setting_name in ["sd_model_checkpoint", "forge_additional_modules"]:
            if shared.opts.disable_weights_auto_swap:
                continue
```

DenOfEquity:
Agreed, thanks for the feedback!

altoiddealer:
This works very good btw, well done sir!

PR #2110, opened 2024-10-18
Make Extension Update Checks Multi-Threaded

Haoming02:
#### What does this PR do?
- Use the `threading` package to make the **Check for updates** parallel

#### Feature
- Now it checks the updates of all Extensions at the same time, instead of checking one by one, thus drastically reducing the waiting time

#### Test
- On my installation, checking 11 Extensions only took `1.1 sec`. And after pressing **Apply and restart UI**, everything updates and works correctly.


Haoming02:
Opened the same PR upstream. If that one gets merged, it would automatically get pulled here later.

Issue #2089, opened 2024-10-16
AttributeError: 'NoneType' object has no attribute 'sd_checkpoint_info'


psydok:
After several requests to change checkpoints this error starts to fall:
```bash
2024-10-16T12:32:26: Done.
2024-10-16T12:32:26: ERROR:    Exception in ASGI application
2024-10-16T12:32:26: Traceback (most recent call last):
2024-10-16T12:32:26:   File "/opt/conda/lib/python3.11/site-packages/uvicorn/protocols/http/h11_impl.py", line 406, in run_asgi
2024-10-16T12:32:26:     result = await app(  # type: ignore[func-returns-value]
2024-10-16T12:32:26:              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-10-16T12:32:26:   File "/opt/conda/lib/python3.11/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
2024-10-16T12:32:26:     return await self.app(scope, receive, send)
2024-10-16T12:32:26:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-10-16T12:32:26:   File "/opt/conda/lib/python3.11/site-packages/fastapi/applications.py", line 1106, in __call__
2024-10-16T12:32:26:     await super().__call__(scope, receive, send)
2024-10-16T12:32:26:   File "/opt/conda/lib/python3.11/site-packages/starlette/applications.py", line 122, in __call__
2024-10-16T12:32:26:     await self.middleware_stack(scope, receive, send)
2024-10-16T12:32:26:   File "/opt/conda/lib/python3.11/site-packages/starlette/middleware/errors.py", line 184, in __call__
2024-10-16T12:32:26:     raise exc
2024-10-16T12:32:26:   File "/opt/conda/lib/python3.11/site-packages/starlette/middleware/errors.py", line 162, in __call__
2024-10-16T12:32:26:     await self.app(scope, receive, _send)
2024-10-16T12:32:26:   File "/opt/conda/lib/python3.11/site-packages/gradio/route_utils.py", line 724, in __call__
2024-10-16T12:32:26:     await self.app(scope, receive, send)
2024-10-16T12:32:26:   File "/opt/conda/lib/python3.11/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
2024-10-16T12:32:26:     raise exc
[...]

PR #2111, opened 2024-10-18
Validate checkpoints in checkpoint_change()

altoiddealer:
I believe this solves https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/2089

_EDITED comment below **in bold text**_

Upon reviewing the relevant code in this issue, I can see that the option `hr_checkpoint_name` was not **100% flawlessly ** validating the checkpoint before applying `checkpoint_change()`.  **in [processing module line 1272](https://github.com/lllyasviel/stable-diffusion-webui-forge/blob/main/modules/processing.py#L1272) I believe that `get_closet_checkpoint_match()` may possibly return an invalid value?**

First pass `sd_checkpoint_model` checkpoints were being validated outside `checkpoint_change()`.

Rather than creating another duplicate validation step specifically for `hr_checkpoint_name` outside the function, simply validate all checkpoints within `checkpoint_change()`

### DIFF
--- a/modules_forge/main_entry.py
+++ b/modules_forge/main_entry.py
@@ -240,6 +240,9 @@ def refresh_model_loading_parameters():
 
 
 def checkpoint_change(ckpt_name:str, save=True, refresh=True):
+    if ckpt_name is not None and ckpt_name not in sd_models.checkpoint_aliases:
+        raise RuntimeError(f"model {ckpt_name!r} not found")
+
     shared.opts.set('sd_model_checkpoint', ckpt_name)
 
     if save:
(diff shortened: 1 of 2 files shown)


altoiddealer:
A part of me wants to mark this as "Draft", because I'm starting to think this does not solve #2089

I'm leaving this open as I think it is still a good idea to validate all requested checkpoint changes in this function

altoiddealer:
Closing this as it only theoretically solves Issues which have not yet been reported

psydok:
I don't have the error anymore yet. It was due to incorrectly set names. Now I'm transmitting correctly. I think this is a useful PR tool for checking the set values.
But I didn't check.

modules_forge/main_entry.py
```diff
@@ -240,6 +240,9 @@ def refresh_model_loading_parameters():
 
 
 def checkpoint_change(ckpt_name:str, save=True, refresh=True):
+    if ckpt_name is not None and ckpt_name not in sd_models.checkpoint_aliases:
+        raise RuntimeError(f"model {ckpt_name!r} not found")
```
psydok, line 244: ValueError is more suitable here
```python
raise ValueError(f"Model {ckpt_name!r} not found")
```

PR #2115, opened 2024-10-19
Fix model prediction detection

### LINKED ISSUE #1109: [Bug]: SD1.5 v-pred checkpoints don't work properly after the latest update
Some of the latest changes are broken 1.5 v-pred models.

Model example: https://civitai.com/models/199059/9th-tail

It has .yaml config with
```
  params:
    parameterization: "v"
```

But for some reason it doesn't work like before.

Tested on clean webui-forge setup.

It creates mess instead of images:

catboxanon:
Addresses #1109

For newer SDXL-based models such as https://civitai.com/models/833294?modelVersionId=962003, authors will need to include a `v_pred` key in the `state_dict` with the model in order for it to be detected correctly.

Related: https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/16567

### DIFF
--- a/backend/loader.py
+++ b/backend/loader.py
@@ -243,6 +243,7 @@ def split_state_dict(sd, additional_state_dicts: list = None):
             sd = replace_state_dict(sd, asd, guess)
 
     guess.clip_target = guess.clip_target(sd)
+    guess.model_type = guess.model_type(sd)
 
     state_dict = {
         guess.unet_target: try_filter_state_dict(sd, guess.unet_key_prefix),
@@ -285,6 +286,15 @@ def forge_loader(sd, additional_state_dicts=None):
                 del state_dicts[component_name]
             if component is not None:
                 huggingface_components[component_name] = component
+
+    # Fix Huggingface prediction type using estimated config detection
+    prediction_types = {
+        'EPS': 'epsilon',
+        'V_PREDICTION': 'v_prediction',
+        'EDM': 'edm',
+    }
+    if 'scheduler' in huggingface_components and hasattr(huggingface_components['scheduler'], 'config') and 'prediction_type' in huggingface_components['scheduler'].config:
+        huggingface_components['scheduler'].config.prediction_type = prediction_types.get(estimated_config.model_type.name, 'epsilon')
 
     for M in possible_models:
         if any(isinstance(estimated_config, x) for x in M.matched_guesses):


rabidcopy:
If models need to have prediction type hard coded into them, what does the end user do for models that don't have this? Example being the model linked.

catboxanon:
SD2 models should hopefully be autodetected by this: https://github.com/lllyasviel/huggingface_guess/blob/84826248b49bb7ca754c73293299c4d4e23a548d/huggingface_guess/model_list.py#L173

The model I linked is now confirmed to have it hardcoded into the next release, which happens via the detection here: https://github.com/lllyasviel/huggingface_guess/blob/84826248b49bb7ca754c73293299c4d4e23a548d/huggingface_guess/model_list.py#L287-L288
(I added a download link for a version with the v_pred key in the related PR linked above)

All of this code actually comes upstream from ComfyUI. https://github.com/comfyanonymous/ComfyUI/blob/7390ff3b1ec2e15017ba4a52d6eaabc4aa4636e3/comfy/supported_models.py

rabidcopy:
And for old v-prediction SD1 models? From my understanding Forge no longer looks for yamls next to the model in Stable-diffusion/. Unless I'm mistaken?

catboxanon:
I can try taking a look at that, but it might be challenging to get working since my guess is that was using the code from upstream A1111 (which can load .yaml configs next to models), and when lllyasviel reworked the entire backend a few months ago, he gutted that out. Forge is completely ComfyUI-based more or less for it's backend now, and I don't believe ComfyUI has support to load .yaml configs next to models like that currently.

rabidcopy:
That's where I'm sort of seeing the disconnect in functionality. reForge for example has this extension that mimics ComfyUI's ModelSamplingDiscrete node. Which allows the user to set prediction type on the fly as well as zsnr and a few other specific architecture-specific sampling options. https://github.com/Panchovix/stable-diffusion-webui-reForge/tree/dev_upstream/extensions-builtin/reForge-advanced_model_sampling
ComfyUI hasn't depended on YAMLs ever since the CheckpointerLoader that allowed specifiying a YAML was deprecated and as such it's been possible to set the prediction for models like this for a while now. ComfyUI does automatically detect prediction for base models and whatnot, but there's been a way to manually set the prediction when the auto-detect doesn't meet user needs when unconventional models arise. I feel like expecting the user to alter their models to be auto-detected is an inconvenient solution compared to just allowing the user to manually set the prediction.

catboxanon:
Oh, I did just realize you're the one that wrote the patch to add it in settings. If you make a PR for that I'll take a second look at it and see about getting it merged. Having it default to "Automatic" with other options overwriting it seems ideal to me.

internetakias:
I tried to add the v_pred key to a SD 1.5 v_pred model but it didn't seem to work. Does the code not look for they key at all in 1.5 models?

catboxanon:
Fixed .yaml loading now. https://github.com/lllyasviel/stable-diffusion-webui-forge/pull/2224

Bocchi-Chan2023:
noobai xl does not produce a normal image. Can't I just load the checkpoints?

Bocchi-Chan2023:
Thank you, it was different from A1111

GromForever:
@catboxanon I followed the instructions exactly - added a .yaml file with the same name as the safetensors file, added the necessary lines to it, restarted forge, but it still ignores v-pred. I pulled all the latest commits from git.
изображение
изображение
What did I do wrong?

PR #2138, opened 2024-10-21
fix text2img api pydantic error

vaskers5:


hwoitke (approved):

altoiddealer:
I tested in the UI as well as in the API:
- using hi res fix with a different checkpoint of the same architecture, and the new `hr_additional_modules`
- not using HR Fix

This fix resolved the error, and the results of using or not using `hr_additional_modules` are behaving expectedly.

---

While correctly using the new `hr_additional_modules`, **I DO get errors if**:
- I try changing from either an XL or SD1.5 model to a Flux model for HR fix
  -XL -> Flux: `mat1 and mat2 shapes cannot be multiplied (5762x16 and 64x3072)`
  - SD1.5 -> Flux: `mat1 and mat2 shapes cannot be multiplied (4032x16 and 64x3072)`
- I try changing from a Flux model to an XL or SD1.5 model for HR Fix
  - Flux > XL: `Given groups=1, weight of size [320, 4, 3, 3], expected input[2, 16, 172, 134] to have 4 channels, but got 16 channels instead`
  - Flux ->SD1.5: `Given groups=1, weight of size [320, 4, 3, 3], expected input[2, 16, 144, 112] to have 4 channels, but got 16 channels instead`

DenOfEquity:
Using latent upscaler?
Latents are not compatible between architectures, even when they have the same number of channels as in 1.5->XL.

altoiddealer:
> Using latent upscaler? Latents are not compatible between architectures, even when they have the same number of channels as in 1.5->XL.

Aha, let me try without Latent...

DenOfEquity:
Thanks for testing, and thanks @vaskers5 for the PR.

PR #2181, opened 2024-10-25
Improve option handling (sd_model_checkpoint / forge_additional_modules)

altoiddealer:
- When comparing inbound setting for `forge_additional_modules` to the stored setting, the lists could have the same values in them but it would trigger `refresh_model_parameters()` if they were in different orders.  Now, the values are sorted when comparing them.
- A wide variety of values are accepted and valid for `sd_model_checkpoint`.  Because of this, when checking if inbound setting for `sd_model_checkpoint` is equivelant to the current setting, these would often not match even if it is for the same model.  I've added some lines that gets a consistent value from a valid checkpoint name/alias/etc, as well as for the current model, and it will compare these.  This will resolve many cases where the same model is being reloaded unnecessarily.

**Edit** - Additionally:
- `hr_checkpoint_name` is now compared with the current checkpoint by the same method, ensuring an apples to apples comparison.
- Updated the handling of `hr_additional_modules` to let the comparison/application be handled within `modules_change()`
- In the `finally` block, explicitly call `main_entry.refresh_model_loading_parameters()` in case the checkpoint does not change (which would skip "refresh")

### DIFF
--- a/modules/processing.py
+++ b/modules/processing.py
@@ -1402,24 +1402,24 @@ class StableDiffusionProcessingTxt2Img(StableDiffusionProcessing):
 
             reload = False
             if 'Use same choices' not in self.hr_additional_modules:
-                if sorted(self.hr_additional_modules) != sorted(fp_additional_modules):
-                    main_entry.modules_change(self.hr_additional_modules, save=False, refresh=False)
+                modules_changed = main_entry.modules_change(self.hr_additional_modules, save=False, refresh=False)
+                if modules_changed:
                     reload = True
 
             if self.hr_checkpoint_name and self.hr_checkpoint_name != 'Use same checkpoint':
-                if self.hr_checkpoint_name != fp_checkpoint:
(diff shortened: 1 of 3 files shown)


psydok:
I tested it. I think it worked. `override_setting` no longer reinitializes same checkpoint. Thanks!

altoiddealer:
The handling of high res checkpoint model needs to be adjusted to utilize this.

Will be updating this PR

altoiddealer:
Made the following changes:

- Revised the checkpoint comparison to use `get_closet_checkpoint_match()`, which is almost identical to how I had set but this is cleaner and slightly more lenient.
- `hr_checkpoint_name` is now compared with the current checkpoint by the same method, ensuring an apples to apples comparison.
- Updated the handling of `hr_additional_modules` to let the comparison/application be handled within `modules_change()`
- In the `finally` block, explicitly call `main_entry.refresh_model_loading_parameters()` in case the checkpoint does not change (which would skip "refresh")

PR #2182, opened 2024-10-25
Update lora.py

sangoi-exe:
Fix for the unmatched keys in DoRA (at least the ones trained in OneTrainer).

From my tests, not having those keys affects the gen.

Error log:
```
[LORA] Loading C:\stable-diffusion-webui\models\Lora\onetrainer\abluble.safetensors for KModel with unmatched keys ['lora_te2_text_projection.alpha', 'lora_te2_text_projection.dora_scale', 'lora_te2_text_projection.lora_down.weight', 'lora_te2_text_projection.lora_up.weight']
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,0 +1,7 @@
+About Gradio 5: will try to upgrade to Gradio 5 at about 2025 March. If failed, then will try again on about 2025 June. relatively positive that we can have Gradio5 before next summer.
+
+2024 Oct 28: A new branch `sd35` is contributed by [#2183](https://github.com/lllyasviel/stable-diffusion-webui-forge/pull/2183) . I will take a look at quants and sampling and transformer's clip-g vs that clip-g rewrite before merging to main ... (Oct 29: okay maybe medium also need to take a look later)
+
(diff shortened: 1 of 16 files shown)


DenOfEquity:
Hi, thanks for your PR.
Looking at `model_lora_keys_clip()` in `packages_3rdparty\comfyui_lora_collection\lora.py`:
it always returns `sdk` ( = `model.state_dict().keys()`) and `key_map`, so the first part of this doesn't do anything?
Then the key checks at the end are already in that function, but currently commented out. So an ideal solution might be to simply restore those lines.
Can you link an example DoRA this is relevant for?

sangoi-exe:
> Hi, thanks for your PR.
> Looking at `model_lora_keys_clip()` in `packages_3rdparty\comfyui_lora_collection\lora.py`:
> it always returns `sdk` ( = `model.state_dict().keys()`) and `key_map`, so the first part of this doesn't do anything?
> Then the key checks at the end are already in that function, but currently commented out. So an ideal solution might be to simply restore those lines.
> Can you link an example DoRA this is relevant for?

Aw, you’re absolutely right. 😅

My knowledge of ForgeUI’s code is pretty superficial, so I didn’t know those checks were already in place.

I didn’t even realize the `comfyui_lora_collection/lora.py` file existed, so I just worked with the first LoRA-related file I found. With the help of GPT o1-preview, I did what I could with the information I had, so it ended up being more of a workaround than a proper solution.

If restoring those commented lines solves the problem, that’s definitely the better way to go. 

As for where this issue shows up, any DoRA model trained on OneTrainer runs into this problem.

sangoi-exe:
btw, as soon as I get some time today, I'll try out that approach of restoring the commented lines to see if it fixes the DoRAs issue

PR #2186, opened 2024-10-26
Any exception is translated in http response

psydok:
Problem: 
-  Requests did not return error text. 
- `api_middleware(self.app)` was disabled. When enabled, it worked only with `--nowebui` argument.

Fix:
- Now at any exception message error is translated in http response (with `--nowebui` and without `--nowebui`)

Mini note:
- Slightly fixed sorting of imports...
- It seems repository hasn't been formatted with [ruff](https://docs.astral.sh/ruff/) (pyprojects.toml) in long time... Maybe we should add [pre-commit](https://pre-commit.com/) so that everyone writes same code?

UPD: rolled back imports in notes

### DIFF
--- a/modules/api/api.py
+++ b/modules/api/api.py
@@ -1,32 +1,35 @@
 import base64
 import io
-import ipaddress
+import os
 import time
-from contextlib import closing
-from io import BytesIO
-from secrets import compare_digest
-from threading import Lock
-from typing import Any, Union, get_args, get_origin
-
-import gradio as gr
-import piexif
-import piexif.helper
-import requests
+import datetime
 import uvicorn
-from fastapi import APIRouter, Depends, FastAPI, Response
-from fastapi.exceptions import HTTPException
+import ipaddress
+import requests
+import gradio as gr
+from threading import Lock
+from io import BytesIO
+from fastapi import APIRouter, Depends, FastAPI, Request, Response
 from fastapi.security import HTTPBasic, HTTPBasicCredentials
-from PIL import PngImagePlugin
+from fastapi.exceptions import HTTPException
+from fastapi.responses import JSONResponse
+from fastapi.encoders import jsonable_encoder
+from secrets import compare_digest
 
 import modules.shared as shared
-from modules import sd_samplers, deepbooru, images, scripts, ui, postprocessing, restart, shared_items, script_callbacks, infotext_utils, sd_models, sd_schedulers
+from modules import sd_samplers, deepbooru, images, scripts, ui, postprocessing, errors, restart, shared_items, script_callbacks, infotext_utils, sd_models, sd_schedulers
 from modules.api import models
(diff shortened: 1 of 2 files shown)


psydok:
@DenOfEquity @altoiddealer I think you've been in charge of repository lately. Please take look!

altoiddealer:
In charge of repository? ha!  Just trying to contribute with relatively simple code, where it is needed.

I'm an amateur, plain and simple.

I can say this, though - you should limit the amount of code you are changing to fulfill the task.  Changing labels for no other reason than to change labels, changing imports for cosmetic purposes - these changes can have unexpected consequences.  For one, since a fair chunk of code derives from A1111, it is helpful to maintain same labelling - later on it is easier to search for and compare code when trying to debug "X works on A1111, why not Forge?"

I'll test this code for awhile and see what happens, but for future reference try solving issues while limiting code changes to minimum required.  If this is all required and I'm overlooking that, then I apologize.

psydok:
@altoiddealer Thank you! That's a good point, I'll return imports.

modules/api/api.py
```diff
@@ -205,7 +205,7 @@ def __init__(self, app: FastAPI, queue_lock: Lock):
         self.router = APIRouter()
         self.app = app
         self.queue_lock = queue_lock
-        #api_middleware(self.app)  # XXX this will have to be fixed
+        #api_middleware(self.app)  # FIXME: (legacy) this will have to be fixed
```
psydok, line 208: I can't decide if I should delete this method and its imports now? With the new version of gradio and fastapi, this method won't work anyway.
Or leave it as it is?

DenOfEquity:
altoiddealer being modest about their very valuable contributions, well worthy of being Collaborator IMO.

modules/api/api.py
DenOfEquity: I'd say leave it for now. If it is certain that it can't be restored then that's an excellent argument to remove, and add a comment in an appropriate place so say why. But no urgency to include that right now unless it's interfering somehow. I'm not an API user, so can't provide useful comments; but this updated PR is much better for reviewability and seems fine to me, so thanks for that. altoiddealer is an API user, so (without wanting to dump responsibility on them) I'll hold off on pressing the button for a short while, for their opinion.

altoiddealer:
Should be able to test this tomorrow

modules/api/api.py
psydok: Ok. Waiting for @altoiddealer

altoiddealer:
This is very good!

I loaded a gguf model and did not load any modules.

Here is the log from my discord bot, where I made txt2img request with current Forge, and one with the PR.

Both requests yielded the same AssertionError in Forge, but the error was only reported from the API response with the PR.

Screenshot 2024-10-27 114446

PR #2192, opened 2024-10-27
Refiner

DenOfEquity:
All seems to work correctly, tested with sd1.5 models with and without hiresfix changing model too; including refiner in both passes; and batch count > 1. But it's long past being too late to test more.
Refiner model should be same architecture - latents are not compatible so changing architecture will give bad results at best and will error out if number of latent channels is different.


DenOfEquity:
Also tested other architectures, LoRAs, controlnet (first pass only?). Second pass uses same additional modules as assigned to first pass - I don't see value in adding another VAE selector to the UI.
The only real change is to use the new model loading functions.

Issue #2196, opened 2024-10-27
'dict' object has no attribute 'get_control' when using Controlnet Tile and MultiDiffusion Integrated with Mixture of Diffusers methon


licyk:
### What Happened?
I using stable diffusion webui forge verison: f2.0.1v1.10.1-previous-589-g41a21f66.

When I using Img2img to upscale image, enabling Controlnet Tile and MultiDiffusion Integrated with Mixture of Diffusers methon, an `'dict' object has no attribute 'get_control'` error has occurred.

### Steps to reproduce the problem
1. open Img2img
2. enter prompt and image
3. enable controlnet integrated and select tile control type, using xinsir-controlnet-union-1.0
4. enable multidiffusion integrated and select mixture of diffusers methon
5. generate

### Console logs
```
Traceback (most recent call last):
  File "D:\Downloads\BaiduNetdiskWorkspace\sd-webui-forge\modules_forge\main_thread.py", line 30, in work
    self.result = self.func(*self.args, **self.kwargs)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Downloads\BaiduNetdiskWorkspace\sd-webui-forge\modules\img2img.py", line 250, in img2img_function
    processed = process_images(p)
                ^^^^^^^^^^^^^^^^^
  File "D:\Downloads\BaiduNetdiskWorkspace\sd-webui-forge\modules\processing.py", line 842, in process_images
    res = process_images_inner(p)
          ^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Downloads\BaiduNetdiskWorkspace\sd-webui-forge\modules\processing.py", line 990, in process_images_inner
    samples_ddim = p.sample(conditioning=p.c, unconditional_conditioning=p.uc, seeds=p.seeds, subseeds=p.subseeds, subseed_strength=p.subseed_strength, prompts=p.prompts)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Downloads\BaiduNetdiskWorkspace\sd-webui-forge\modules\processing.py", line 1865, in sample
[...]

PR #2199, opened 2024-10-28
fix multidiffusion integrated

licyk:
fix: https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/2196


DenOfEquity:
Thank you for the fix.

Issue #2166, opened 2024-10-24
"'NoneType' object has no attribute 'sd_checkpoint_info'", if prior was "You do not have CLIP state dict!"


dermesut:
once i get "you don't have state dict", i can't generate an image with the sd model that is set, even if i complete the state dict, due to "'NoneType' object has no attribute 'sd_checkpoint_info'"
only after i change the sd model to something else, generate an image with that one, and then change back to my original sd model, i can generate an image again.

this is the process, that lets me reproduce that issue:

1) start of server

```
Python 3.10.6 (tags/v3.10.6:9c7b4bd, Aug  1 2022, 21:53:49) [MSC v.1932 64 bit (AMD64)]
Version: f2.0.1v1.10.1-previous-584-g9a698e26
Commit hash: 9a698e26d6744de24d05568c9938a52694dbb3f0
Faceswaplab : Use GPU requirements
Checking faceswaplab requirements
Install protobuf>=3.20.2
Installing sd-webui-faceswaplab requirement: protobuf>=3.20.2
1.1965869999985443
CUDA 12.1
Launching Web UI with arguments: --api --port 7861 --gpu-device-id 1 --wildcards-dir 'E:\ai\_gh_repos\sd.webui\webui\extensions\stable-diffusion-webui-wildcards\wildcards' --forge-ref-a1111-home 'E:\ai\_gh_repos\sd.webui\webui' --text-encoder-dir 'E:\ai\_gh_repos\sd.webui\webui\models\text_encoder' --xformers --cuda-malloc --cuda-stream --pin-shared-memory --ckpt-dir 'E:\ai\_gh_repos\sd.webui\webui\models\Stable-diffusion' --vae-dir 'E:\ai\_gh_repos\sd.webui\webui\models\VAE' --hypernetwork-dir 'E:\ai\_gh_repos\sd.webui\webui\models\hypernetworks' --embeddings-dir 'E:\ai\_gh_repos\sd.webui\webui\embeddings' --lora-dir 'E:\ai\_gh_repos\sd.webui\webui\models\lora' --controlnet-dir 'E:\ai\_gh_repos\sd.webui\webui\models\ControlNet' --controlnet-preprocessor-models-dir 'E:\ai\_gh_repos\sd.webui\webui\extensions\sd-webui-controlnet\annotator\downloads'
Set device to: 1
Using cudaMallocAsync backend.
Total VRAM 11264 MB, total RAM 130956 MB
[...]

PR #2215, opened 2024-10-29
Minimal preload before dump current model data

altoiddealer:
Requesting review from **Illyasviel** 

Currently, all loaded model data is trashed and garbage collected before Forge gathers configs / components / etc.
If an AssertionError is raised after this due to a missing module such as a required text encoder, or potentially other future compatibility issues we want to check for - this throws a monkey wrench into whatever model was most recently loaded (https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/2166).

What this PR does is performs the minimum steps needed to verify model/modules before the loaded model data gets trashed, subsequently preventing unnecessary computation and headaches due to a missing required text encoder, etc.

The new `forge_preloader()` returns the few items of any significance it collects in the process for the `forge_loader()` to use.

### DIFF
--- a/backend/loader.py
+++ b/backend/loader.py
@@ -30,6 +30,24 @@ logging.getLogger("diffusers").setLevel(logging.ERROR)
 dir_path = os.path.dirname(__file__)
 
 
+def check_huggingface_component(component_name:str, cls_name:str, state_dict):
+    check_sd = False
+    comp_str = 'model'
+
+    if cls_name == 'AutoencoderKL':
+        check_sd = True
+        comp_str = 'VAE'
+    elif component_name.startswith('text_encoder') and cls_name in ['CLIPTextModel', 'CLIPTextModelWithProjection']:
+        check_sd = True
+        comp_str = 'CLIP'
+    elif cls_name == 'T5EncoderModel':
+        check_sd = True
+        comp_str = 'T5'
+
+    if check_sd and (not isinstance(state_dict, dict) or len(state_dict) <= 16):
+        raise AssertionError(f'You do not have {comp_str} state dict!')
+
+
 def load_huggingface_component(guess, component_name, lib_name, cls_name, repo_path, state_dict):
     config_path = os.path.join(repo_path, component_name)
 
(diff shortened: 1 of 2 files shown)


backend/loader.py
```diff
@@ -30,6 +30,24 @@
 dir_path = os.path.dirname(__file__)
 
 
+def check_huggingface_component(component_name:str, cls_name:str, state_dict):
```
psydok, line 33: in near future, I may be engaged in restoring sd_checkpoints_limit (but not yet a fact). then you will have to work with these modules. so I decided to take a look at the PR.

Spaces? Typehint?
```suggestion
def _check_huggingface_component(component_name: str, cls_name: str, state_dict: dict[str, Any] | list[?]):
```

backend/loader.py
```diff
@@ -30,6 +30,24 @@
...
+        check_sd = True
+        comp_str = 'VAE'
+    elif component_name.startswith('text_encoder') and cls_name in ['CLIPTextModel', 'CLIPTextModelWithProjection']:
+        check_sd = True
+        comp_str = 'CLIP'
+    elif cls_name == 'T5EncoderModel':
+        check_sd = True
+        comp_str = 'T5'
+
+    if check_sd and (not isinstance(state_dict, dict) or len(state_dict) <= 16):
```
psydok, line 47: could you please explain this magic number 16? Why 16?

altoiddealer: @psydok to answer both questions, the `state_dict` variable and that specific condition, were both lifted from `load_huggingface_component()` which is called from `forge_loader()`.

These are Illyasviel code - I simply relocated this conditional check.  With this PR it is now checking the condition before trashing the currently loaded model data, so if an exception is raised the current model data does not get unloaded.

altoiddealer: Aside from the context clues of this condition, I otherwise have no idea what the expected structure of `state_dict` is supposed to be.  There is even less type hinting where I lifted the code from `load_huggingface_component()`

altoiddealer:
Illyasviel - this PR solves a very annoying, and very easily reproduceable bug.

Reproduce bug: simply fail to include a text encoder or other module that is not baked in to the SD model.

Bug: SD model becomes unusable until you switch to another model, successfully generate an image - THEN you can switch back to your original model and try again.  Otherwise it just yields `AttributeError: 'NoneType' object has no attribute 'sd_checkpoint_info'` endlessly forever.

backend/loader.py
psydok: Okay, I got it! Thank you very much!

PR #2240, opened 2024-11-01
Restore lines for DoRA TE keys fix

sangoi-exe:
As stated by [DenOfEquity](https://github.com/DenOfEquity), the best approach would be to restore these lines instead of doing a workaround in `backend>patcher>lora.py`

>Restoring lines in the `model_lora_keys_clip` function from `packages_3rdparty>comfyui_lora_collection>lora.py` to fix the issue where some TE keys aren’t recognized when loading a DoRA.

### DIFF
--- a/packages_3rdparty/comfyui_lora_collection/lora.py
+++ b/packages_3rdparty/comfyui_lora_collection/lora.py
@@ -245,15 +245,15 @@ def model_lora_keys_clip(model, key_map={}):
                 lora_key = "lora_te2_{}".format(l_key.replace(".", "_"))#OneTrainer Flux lora, by Forge
                 key_map[lora_key] = k
                 #####
-            elif k.startswith("hydit_clip.transformer.bert."): #HunyuanDiT Lora
-                l_key = k[len("hydit_clip.transformer.bert."):-len(".weight")]
-                lora_key = "lora_te1_{}".format(l_key.replace(".", "_"))
-                key_map[lora_key] = k
+    #         elif k.startswith("hydit_clip.transformer.bert."): #HunyuanDiT Lora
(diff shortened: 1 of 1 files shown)


lllyasviel:
give a lora link that you want to support

sangoi-exe:
> give a lora link that you want to support

[Shiftymine.cyb22](https://drive.google.com/file/d/1gjJGU_iZ_lIUBqM6UdqvLifnvhJRBG2-/view?usp=sharing)

Trained on cyberrealisticXL_v22.

Careful, this DoRA might create nsfw content and well-endowed bust. No trigger word needed.

DenOfEquity:
Confirmed that the linked LoRA shows unmatched keys when loaded, and results after correction are different (limited testing, but seem better). Strictly, this PR uncomments more than necessary - but the SD3 and Flux lines could be relevant anyway / soon, and the HyDit and Cascade lines shouldn't cause any issues AFAIK. (Already some handling of Cascade loras earlier in function.)

sangoi-exe:
> Confirmed that the linked LoRA shows unmatched keys when loaded, and results after correction are different (limited testing, but seem better). Strictly, this PR uncomments more than necessary - but the SD3 and Flux lines could be relevant anyway / soon, and the HyDit and Cascade lines shouldn't cause any issues AFAIK. (Already some handling of Cascade loras earlier in function.)

I think that's it, maybe

DenOfEquity:
Looking more at the T5 section, it seems potentially broken with SD3(.5) (depending on how keys are named, which I'm unsure about) - writing both te2 and te3 keys might overwrite a previously set te2 key. May need to include a check for clip_g. But that can be done later, if it is necessary.

PR #2253, opened 2024-11-04
Add safe getter for override setting of processing

dalapenko:
After my previous fix some users report problems with processing https://github.com/lllyasviel/stable-diffusion-webui-forge/pull/1952#issuecomment-2453782721

Root cause is initialize of processing override_settings  property like null. One of solution it's check in a place where i'm use, but override_settings uses without checking to null more widely in project, that in one place. 

In this PR i'm change init of this property, now it's not null ever

### DIFF
--- a/modules/images.py
+++ b/modules/images.py
@@ -642,7 +642,7 @@ def save_image(image, path, basename, seed=None, prompt=None, extension='png', i
             Additional PNG info. `existing_info == {pngsectionname: info, ...}`
         no_prompt:
             TODO I don't know its meaning.
-        p (`StableDiffusionProcessing`)
+        p (`StableDiffusionProcessing` or `Processing`)
         forced_filename (`str`):
             If specified, `basename` and filename pattern will be ignored.
         save_to_dirs (bool):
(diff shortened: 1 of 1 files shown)


dalapenko:
so.. no, it's not a problem from user issues. override_settings already set with default dict in post init. will research

dalapenko:
@rushuna86 @wensleyoliv

can you check fix at my fork? i can't reproduce crash from your report, but seems like fix it
https://github.com/dalapenko/stable-diffusion-webui-forge/tree/main

DenOfEquity:
Error is when XYZ (and probably aDetailer) sends the `Processed` object, which doesn't have `override_settings`, in place of a `Processing` object when saving an image.
At first I couldn't replicate the issue either, but then I noticed that `file_decoration` is set firstly based on `opts.samples_filename_pattern` and only if that isn't set then it uses `p.override_settings.get("samples_filename_pattern")` (then finally the fallback name). *The override should be checked first.* Because I had a save filename pattern set (*Settings > Saving images/grids > Images filename pattern*), the override was never checked and so the error couldn't occur.
A simpler (IMO) solution, from line 673 of *images.py*:
```
    if forced_filename is None:
        if short_filename or seed is None:
            file_decoration = ""
        elif hasattr(p, 'override_settings'):
            file_decoration = p.override_settings.get("samples_filename_pattern")
        else:
            file_decoration = None
        
        if file_decoration is None:
            file_decoration = opts.samples_filename_pattern or ("[seed]" if opts.save_to_dirs else "[seed]-[prompt_spaces]")
            
        file_decoration = namegen.apply(file_decoration) + suffix
```

dalapenko:
@DenOfEquity Ok, it's works similary - i have no objection. Made it like you say

Issue #2289, opened 2024-11-09
SelfAttentionGuidance Integrated (SD 1.x, SD 2.x, SDXL)


armpitranger:
When I use it, the following error occurs

Traceback  (most recent call last):
  File "/content/stable-diffusion-webui-forge/modules_forge/main_thread.py", line 30, in work
    self.result = self.func(*self.args, **self.kwargs)
  File "/content/stable-diffusion-webui-forge/modules/txt2img.py", line 124, in txt2img_function
    processed = processing.process_images(p)
  File "/content/stable-diffusion-webui-forge/modules/processing.py", line 842, in process_images
    res = process_images_inner(p)
  File "/content/stable-diffusion-webui-forge/modules/processing.py", line 990, in process_images_inner
    samples_ddim = p.sample(conditioning=p.c, unconditional_conditioning=p.uc, seeds=p.seeds, subseeds=p.subseeds, subseed_strength=p.subseed_strength, prompts=p.prompts)
  File "/content/stable-diffusion-webui-forge/modules/processing.py", line 1387, in sample
    samples = self.sampler.sample(self, x, conditioning, unconditional_conditioning, image_conditioning=self.txt2img_image_conditioning(x))
  File "/content/stable-diffusion-webui-forge/modules/sd_samplers_kdiffusion.py", line 238, in sample
    samples = self.launch_sampling(steps, lambda: self.func(self.model_wrap_cfg, x, extra_args=self.sampler_extra_args, disable=False, callback=self.callback_state, **extra_params_kwargs))
  File "/content/stable-diffusion-webui-forge/modules/sd_samplers_common.py", line 272, in launch_sampling
    return func()
  File "/content/stable-diffusion-webui-forge/modules/sd_samplers_kdiffusion.py", line 238, in <lambda>
    samples = self.launch_sampling(steps, lambda: self.func(self.model_wrap_cfg, x, extra_args=self.sampler_extra_args, disable=False, callback=self.callback_state, **extra_params_kwargs))
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
  File "/content/stable-diffusion-webui-forge/k_diffusion/sampling.py", line 146, in sample_euler_ancestral
    denoised = model(x, sigmas[i] * s_in, **extra_args)
[...]

PR #2294, opened 2024-11-09
SAG: try old mask reshape calculation first

DenOfEquity:
#2289
crowd testing found situations where the new calculation fails (ex: 1024*584), rounding up when it shouldn't
now: try the old calculation first, use new calculation only if the old way fails

### DIFF
--- a/extensions-builtin/sd_forge_sag/scripts/forge_sag.py
+++ b/extensions-builtin/sd_forge_sag/scripts/forge_sag.py
@@ -62,13 +62,19 @@ def create_blur_map(x0, attn, sigma=3.0, threshold=1.0):
     attn = attn.reshape(b, -1, hw1, hw2)
     # Global Average Pool
     mask = attn.mean(1, keepdim=False).sum(1, keepdim=False) > threshold
-
-    f = float(lh) / float(lw)
-    fh = f ** 0.5
-    fw = (1/f) ** 0.5
-    S = mask.size(1) ** 0.5
-    w = int(0.5 + S * fw)
-    h = int(0.5 + S * fh)
+    ratio = 2**(math.ceil(math.sqrt(lh * lw / hw1)) - 1).bit_length()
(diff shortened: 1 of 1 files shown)


DenOfEquity:
more testing:
pushing the aspect ratio to unusual extents can cause errors with the new calculation.
I'd previously tested around typical resolutions including down to around 1024x640, including runs of [w, w+8, w+16, ...], but didn't think that the calculations would break down at less typical resolutions / aspect ratios. Testing also complicated as resolutions that work in sd1.5 might fail with sdxl. The basic problem is the w:h ratio of the shrunk attention only approximates the aspect ratio of the latents. Rounding up is often, but not always, correct. And sometimes rounding down still gives a too-large result. So a simple, single formula solution doesn't seem possible.
current code which I haven't been able to break - calculation to get to the right area then a mini-search, tested out to stupid resolutions like 1016x64, both SAG only and also with low and high Kohya HRFix downscale factors:
```
    # Global Average Pool
    mask = attn.mean(1, keepdim=False).sum(1, keepdim=False) > threshold
    target = mask.size(1)
    ratio = lw / lh
    
    h = int((target / ratio) ** 0.5)
    h_candidates = [h, h-1, h+1]

    for candidate in h_candidates:
        if candidate > 0 and candidate*(target // candidate) == target:
            h = candidate
            w = target // h
            break
```

DenOfEquity:
The approach above 'works' for various interpretations of that word. It always finds a matching shape, but relatively rarely (possibly only with unreasonable image sizes) doesn't find the ideal match. Short of an exhaustive brute force search, I don't think there is a way to reach 100% success.
Instead, it is much easier to cheat.
Modify the Kohya HRFix / Deep Shrink extension to save the shape it resizes to, store in some *shared* location. Then simple calculations in the SAG extension. This also opens the possibility of different downscale factors for width/height.
#2304

PR #2313, opened 2024-11-13
Add a flag to skip initializing google_blockly

Haoming02:
This PR adds a new **cmd_args**, `--skip-google-blockly`, that allows users to skip loading the **Google Blockly** modules, which only adds like 2 samplers anyway afaik

Simply add the flag after the **COMMANDLINE_ARGS** in `webui-user.bat` like other args

<hr>

I have no comment on the *"allegations."* But loading the `google_blockly` sure takes its sweet time every single launch **and** `Reload UI`...

<p align="center">
<br>
<code>before</code>
</p>

<p align="center">
<br>
<code>after</code>
</p>


IrineSistiana:
+1
initialize google blockly	32.263
import torch	10.187

I've already disabled blockly for a while, since I only use sdxl. No problem for me so far.

DenOfEquity:
Seems like a good addition for people who don't care about Flux. There's also `icl_v2.pyz` in the Forge Blockly repo, but that would be Flux based too.
Certainly speeds up loading significantly.
Thanks for the PR.

PR #2349, opened 2024-11-20
Add source_text_component to Results paste buttons

wcole3:
Adds a source_text_component to the Results panel paste buttons' ParamBinding so that the `callable` fields are correctly pasted.

This allows sampling method, schedule type, styles, etc (all fields that are defined mostly as lambdas) to be correctly set when pasting.

There might be a gotcha; the default config value for `send_seed` is True, but prior to this fix, the seed was not getting sent regardless of the value.  With the fix, people would need to change the send_seed value to maintain their current workflow.

### DIFF
--- a/modules/ui_common.py
+++ b/modules/ui_common.py
@@ -15,17 +15,16 @@ import modules.infotext_utils as parameters_copypaste
 folder_symbol = '\U0001f4c2'  # 📂
 refresh_symbol = '\U0001f504'  # 🔄
 
-
 def update_generation_info(generation_info, html_info, img_index):
     try:
         generation_info = json.loads(generation_info)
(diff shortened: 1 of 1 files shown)


DenOfEquity:
Definitely a good improvement!
Suggested changes for line 265, minimal difference in practice but seems more correct:
`res.infotext_plaintext = gr.HTML(visible=False, elem_id=f'infotext_plaintext_{tabname}', elem_classes="infotext")`

An inconsistency in the original, that seems safe to correct at the same time:
* The `html_log` element, line 266, should have `elem_classes='html-log'`, same as line 217.

wcole3:
Good catch, Should be consistent now.

Issue #2351, opened 2024-11-20
Duplicated Sampling method when Reload


danilomaiaweb:
Please see the image...

![image](https://github.com/user-attachments/assets/d18190e2-240a-40b8-85fb-47c720db611e)

PR #2354, opened 2024-11-21
Don't duplicate samplers when adding

wcole3:
Simple check to avoid continually adding samplers to list.

As far as I know, the only samplers that add through this method are the Flux Realistic ones, but you can notice them duplicated for every UI reload.

#2351


DenOfEquity:
Nice one; I'd mistakenly assumed this was an issue that had to be solved inside the blockly code.
Lines 68 and 69 are better inside the conditional too?

wcole3:
That's probably a good idea, I'll test it out tonight.

wcole3:
No unexpected misbehavior 😁

Issue #2359, opened 2024-11-21
Image seed pasted when sending to img2img or inpainting tab


Gushousekai195:
Whenever I send a generated image to img2img or inpainting using the buttons underneath the result image, the exact seed of the generation gets sent, too. This did not happen in previous iterations of the UI and is not desirable.

Image generation parameters also change when sending the image, especially when sending them back into the img2img or inpainting tabs.

wcole3:
See discussion in linked pull requests.  After it's merged the `send_seed` option in config.json should control the Seed paste behavior.

If you feel there are other generation parameters that should have configurable behavior describe them here.

Gushousekai195:
How do I git pull a pull request again?

wcole3:
```
git fetch origin pull/2360/head:respect-send-seed
git switch respect-send-seed
```

ChooChoo38:
In addition to sending the seed, it's also turning off scripts and appending LoRA's to the prompt in a case where you wouldn't want it too. Examples:

1) I have the Ultimate SD Upscale script turned on in the img2img tab. Clicking "Send image and generation parameters to img2img tab" sets Scripts to none, disabling it. I have to re-enable it every time.

2) I add a LoRA via the "Add network to prompt" dropdown. This appends it to the prompt when you generate so you don't have to manually add it to the prompt text box yourself. However, when I send the image to the img2img tab, it appends to LoRA text to the prompt. It shouldn't do that in this case, given you've already got it enabled in the dropdown.

ChooChoo38:
Unsure if this is related, so I'll post it here instead of the request.

In addition to sending the seed, it's also turning off scripts and appending LoRA's to the prompt in a case where you might not want it to. Examples:

1. I have the Ultimate SD Upscale script turned on in the img2img tab. Going to txt2img and clicking "Send image and generation parameters to img2img tab" sets "Script" to none in the img2img tab, disabling it. I have to re-enable it every time.

2. I add a LoRA via the "Add network to prompt" dropdown. This appends it to the prompt when you generate so you don't have to manually add it to the prompt text box yourself. However, when I send the image to the img2img tab, it automatically adds the LoRA text to the prompt itself. It shouldn't do that in this case, given I've already got it enabled in the dropdown.

wcole3:
Thanks, almost certainly related. For 2) when pasting the Lora gets added to the prompt AND is still selected in the drop-down? Or is it just added to the prompt. 

[...]

PR #2360, opened 2024-11-22
Respect send seed

wcole3:
Correction to #2349, fields to paste needs to be filtered for Seed wrt `send_seed` config option.  Added logic for the `send_size` option too, but it seems less likely that someone wouldn't want that to be True.

Might be worth a discussion if the `send_seed` option should be False by default, but idk what the impact of changing the config.json and associated `shared_option` code would be.  Thoughts very welcome.

#2359

### DIFF
--- a/modules/infotext_utils.py
+++ b/modules/infotext_utils.py
@@ -174,8 +174,9 @@ def connect_paste_params_buttons():
             )
 
         if binding.source_text_component is not None and fields is not None:
-            # get all the fields except seed if option set
-            filtered_fields = [f for f in fields if f[1] != "Seed"] if not shared.opts.send_seed else fields
+            # get all the fields except those the user has configured
+            dont_paste_fileds = (["Seed"] if not shared.opts.send_seed else []) + (["Size-1", "Size-2"] if not shared.opts.send_size else [])
+            filtered_fields = [f for f in fields if f[1] not in dont_paste_fileds]
             connect_paste(binding.paste_button, filtered_fields, binding.source_text_component, override_settings_component, binding.tabname)
 
         if binding.source_tabname is not None and fields is not None:
--- a/modules/infotext_utils.py
+++ b/modules/infotext_utils.py
@@ -174,7 +174,9 @@ def connect_paste_params_buttons():
             )
 
         if binding.source_text_component is not None and fields is not None:
-            connect_paste(binding.paste_button, fields, binding.source_text_component, override_settings_component, binding.tabname)
+            # get all the fields except seed if option set
+            filtered_fields = [f for f in fields if f[1] != "Seed"] if not shared.opts.send_seed else fields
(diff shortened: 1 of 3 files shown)


DenOfEquity:
These options will also be used by PNG Info send to, and the paste button. So they should not default to False. (These are the places `source_text_component` was set up previously.)
The setup for this pasting is done on UI creation, so the *send_seed* and *send_size* options (in *shared_options.py*) should have `.needs_reload_ui()` added.
Typo in `dont_paste_fileds`.
There's some odd behaviour after repeated sends, width/height can alternate between values for a cycle or two. (send, change width/height, send again). Changing line 182 to `elif` seems to fix - avoids repeated setting of components.

DenOfEquity:
Thanks.
The elif path could be used by extensions, safer to leave it.
What is Results tab?

wcole3:
Understood, makes sense.  Sorry "Results" panel; i,e. `OutputPanel`

wcole3:
Found another undesirable behavior, the information pasted always comes from the last generated image, which might not correspond to the image a user is trying to paste.  You can observe this by generating an image, pasting to img2img (should work), changing some of the generation parameters, running img2img, and then trying to paste the original image generated in txt2img again.  The information that gets sent is from the last img2img generation.

I'm pretty sure it's due to how I hooked into the button that updates the generation_info, I need to find a way to get that info from the image (which does still get pasted correctly) directly.  If I can't figure it out tomorrow, I'll put open a revert PR.

wcole3:
I believe that last commit addresses the issue, but keeping this in draft to do more testing tomorrow

ChooChoo38:
In addition to sending the seed, it's also turning off scripts and appending LoRA's to the prompt in a case where you might not want it to. Examples:

1) I have the Ultimate SD Upscale script turned on in the img2img tab. Going to txt2img and clicking "Send image and generation parameters to img2img tab" sets "Script" to none in the img2img tab, disabling it. I have to re-enable it every time.

2) I add a LoRA via the "Add network to prompt" dropdown. This appends it to the prompt when you generate so you don't have to manually add it to the prompt text box yourself. However, when I send the image to the img2img tab, it automatically adds the LoRA text to the prompt itself. It shouldn't do that in this case, given I've already got it enabled in the dropdown.

DenOfEquity:
The more I think about this (+ previous PR), the more I think it's not ideal. It copies all settings, including those of extensions, so the likelihood of unwanted overwriting is high. It also always copies settings between img2img <-> inpaint, when that is probably unnecessary (already set) or especially likely to be undesirable (user has adjusted).
If full settings transfer is wanted, it can be done as a two stage process, via PNG Info, anyway. So currently I think reverting the buttons to have the original lighter touch is best. (doesn't need a new PR btw, I can Revert the previous one.)

wcole3:
Sounds good to me.

Issue #177, opened 2024-02-10
[Bug]: Live preview not working with fp8. 


Necro-mancer:
### Checklist

- [x] The issue exists after disabling all extensions
- [x] The issue exists on a clean installation of webui
- [ ] The issue is caused by an extension, but I believe it is caused by a bug in the webui
- [x] The issue exists in the current version of the webui
- [ ] The issue has not been reported before recently
- [ ] The issue has been reported before but has not been fixed yet

### What happened?

RTX 3070 8gb, windows 11, latest drivers.

Fp8 doesn't work. Is it a mistake on my end or is it something that has not been implemented yet?

Toggling the option in settings has no effect on vram usage, even after restarting webui and console.
I tried the fp8-unet commandline arguments but it gives out cuda errors. This is the exact error: RuntimeError: "div_true_cuda" not implemented for 'Float8_e4m3fn' (same for float8_e5)
I know that forge inherently uses less vram than auto1111 at default settings. But with fp8 enabled in auto1111, i can do much more with sdxl without overflowing into shared vram (and generation slowing to a crawl).

### Steps to reproduce the problem

Enable fp8 from settings.
Generate image.

### What should have happened?

Use less vram

### What browsers do you use to access the UI ?

Mozilla Firefox

### Sysinfo

[sysinfo-2024-02-10-19-37.json](https://github.com/lllyasviel/stable-diffusion-webui-forge/files/14230302/sysinfo-2024-02-10-19-37.json)

### Console logs

```Shell
venv "C:\Users\Admin\stable-diffusion-webui-forge\venv\Scripts\Python.exe"
Python 3.10.6 (tags/v3.10.6:9c7b4bd, Aug  1 2022, 21:53:49) [MSC v.1932 64 bit (AMD64)]
Version: f0.0.12-latest-110-g15bb49e7
Commit hash: 15bb49e761e837c0a3463a736762d11941ea69f7
Faceswaplab : Use GPU requirements
Checking faceswaplab requirements
0.008895099999790546
Launching Web UI with arguments:
Total VRAM 8192 MB, total RAM 32689 MB
Set vram state to: NORMAL_VRAM
Device: cuda:0 NVIDIA GeForce RTX 3070 : native
VAE dtype: torch.bfloat16
Using pytorch cross attention
ControlNet preprocessor location: C:\Users\Admin\stable-diffusion-webui-forge\models\ControlNetPreprocessor
[...]

PR #2412, opened 2024-12-04
Fix LivePreview not working with fp8 UNet

Haoming02:
### Simple Description

Many components in Forge, incluing `sd_vae_taesd`, uses `devices.dtype` as the type, and `devices.dtype` is simply set to `memory_management.unet_dtype()`, which would be `fp16` by default if supported *(`fp32` otherwise for older systems)*.

The issue occurs when the user sets the `unet_dtype` to `fp8` instead, either by using the `--unet-in-fp8-e4m3fn` arg, or by setting the `FP8 weight` option supposedly *(which doesn't actually work in Forge)*.

If the `Live preview method` is set to `TAESD` and `dtype` is set to `float8_e4m3fn`, the live preview breaks.

### Summary of Changes

Keep the `dtype` variable at `torch.float32` if the system does not support half precision; otherwise keep it at `torch.float16`

### Related Issue
[#177](https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/177), still persists in the post-Gradio4 Forge


DenOfEquity:
Works for me.
Also fixes issue with inpainting when using same command line argument.

Issue #2422, opened 2024-12-06
Emphasis Mode does not work


Dagdgg:
If you change Emphasis Mode to any mode, nothing actually happens. It will not ignore (emphasis:1.2) prompts, it will not ignore them and treat them as pure text, it will not use normalization. All 4 options are ignored and defaulted to be treated as "Original".

You can also see when loading in an image with "PNG info" that an image generated with a different mode set has the "original" mode imported. Implying that when you change modes by pressing the different gradio buttons, it does not actually save it to the config. Yet it is interpreted as you having chosen a different mode as it does add the override.

This is mainly an issue for SDXL based models who generally prefer the emphasis to be normalized. If they are done the original way it can overload the model quickly and result in garbled mess instead of a functional image with an actual emphasis on whatever was emphasised.

[stable-diffusion-webui-forge](https://github.com/lllyasviel/stable-diffusion-webui-forge/tree/main)/[backend](https://github.com/lllyasviel/stable-diffusion-webui-forge/tree/main/backend)/[text_processing](https://github.com/lllyasviel/stable-diffusion-webui-forge/tree/main/backend/text_processing)/emphasis.py

![Emphasis-Mode](https://github.com/user-attachments/assets/7e8ac0a1-8889-4fa7-971b-d414587d17ec)

PR #2423, opened 2024-12-07
Apply prompt emphasis mode changes immediately

DenOfEquity:
Currently changes to prompt emphasis mode apply only on model load, this PR checks the setting at the start of text encoding too.
Also added handling for mode *'None'* (literal interpretation), which currently gives results identical to mode *'Ignore'*.
#2422

### to do? ###
These changes mean the text processing engine parameter `emphasis_name` is unused: perhaps should be removed along with edits to the models in `backend\diffusion_engine` so they don't send it; and then `dynamic_args['emphasis_name']` could be removed too. Possibly better in a separate PR, to minimise changes in this one and facilitate testing. <sub>Or maybe I'm being lazy tonight.</sub>

### DIFF
--- a/backend/text_processing/classic_engine.py
+++ b/backend/text_processing/classic_engine.py
@@ -6,6 +6,8 @@ from backend.text_processing import parsing, emphasis
 from backend.text_processing.textual_inversion import EmbeddingDatabase
 from backend import memory_management
 
+from modules.shared import opts
+
 
 PromptChunkFix = namedtuple('PromptChunkFix', ['offset', 'embedding'])
 last_extra_generation_params = {}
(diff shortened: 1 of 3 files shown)


Dagdgg:
Copied the changes to the relevant \backend\text_processing files, on the positive side, nothing changed negatively. On the negative side, it seems to have made no difference.

The comment about it loading the chosen emphasis mode on load is very interesting though. I tried now to change mode first then change the model. Generate an image with a token emphasised. Swap back to the other model with the same mode still on, generate, then switch mode, load up the other model and repeat until all 4 different modes have generated a result.

The results were still identical between "None" and "Ignore" so there might be some mistake in how you tried to make it treat all characters literally, or there might be some other code that overrides it. "Original" and "No Norm" however did in fact change results.

Hope this does gets solved without having to reload models in the future, but thanks for the insightful comment, wouldn't have thought to try it out this way, now I at least have a way to change modes. Cheers!

DenOfEquity:
Thanks for update and testing. If you loaded parameters from a previous generation, you may have had an override applied for the setting. I made exactly that mistake just now when retesting.

PR #2435, opened 2024-12-08
just to make it easier for new linux users to find the dark mode option

kairin:
took me an evening to figure out how "./webui.sh" launches the app in a smarter way instead of "python launch.py" 

also to make dark mode easier to figure out. the first example did not made sense to me because i was not able to see the relation between the webui-user.sh and webui.sh

now that i realised it... i hope this minor change can help other future new users to better relate the two files.

i.e. update webui-user.sh to include arguments. then launch using ./webui.sh


UnconnectedBedna:
WDYM?

if the file exists, `webui-user.sh` is sourced in `webui.sh`

https://github.com/lllyasviel/stable-diffusion-webui-forge/blob/f53307881bfd824dbdce6ac0d4bba04d9a74ab36/webui.sh#L22

PR #2489, opened 2024-12-22
add batch (directory) processing

DenOfEquity:
main work by @nitinmukesh (https://github.com/lllyasviel/stable-diffusion-webui-forge/discussions/2458)

added option to always save as PNG
added option to always save flat (apply mask instead of saving with mask channel)
fix for jpg/jpeg (can't save with mask) (if not save flat, save as PNG)
minor UI tweaks

Co-authored-by: nitinmukesh <nitinmukesh@users.noreply.github.com>

### DIFF
--- a/extensions-builtin/forge_space_birefnet/forge_app.py
+++ b/extensions-builtin/forge_space_birefnet/forge_app.py
@@ -1,3 +1,4 @@
+
 import spaces
 import os
 import gradio as gr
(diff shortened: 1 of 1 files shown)


softtaco1:
Thank you very much for this!! This was a feature I’d actually requested for as well, it will make training Loras much easier.

nitinmukesh:
@DenOfEquity 

Thank you very much for creating a PR. I am not a developer so take help from ChatGPT and Claude to make code work as per my need. I just explain the logic and these tools help with syntax. :)
Good to see it might help other as well as it is now integrated.

DenOfEquity:
You're welcome, thanks for sharing. Definitely a useful addition.

PR #2497, opened 2024-12-25
add ldm_patcher folder,in order to feat lots of extensions

spawner1145:


Haoming02:
Or tell those Extensions to update...

Most of these files already exist here: https://github.com/lllyasviel/stable-diffusion-webui-forge/tree/main/backend

PR #2595, opened 2025-01-29
changed to rocm6.2

VeryUnstableRice:


YyLlBbYy:
results in increase of VRAM usage, triggering the need for tiled VAE for image generation at resolutions that worked just fine with the rocm5.7 stack.

hapwizard:
This is necessary for RX7800 XT (Navi 32) cards.  Rocm5.7 does not recognize the card, forces a cuda check and won't launch. Skipping the cuda check just leads to more errors. This fix to install rocm6.2 allows webui to install and run without issue on linux+rocm systems with newer AMD cards.

PR #2183, opened 2024-10-26
sd3.5 integration (naive)

graemeniedermayer:
Minimal integration of SD3.5. Requires update to huggingface_guess https://github.com/lllyasviel/huggingface_guess/pull/1 

There's a a few hacky things. In particular in `backend/loader.py` there's a few things that could be structured better.

example

All text models are necessary for this implementation. Works with 8vram cards.

PR #2642, opened 2025-02-10
increased support for custom CLIPs

DenOfEquity:
custom CLIPs can be applied to sd1.5, sdxl, (sd3, untested) by including them in the VAE / Text Encoder dropdown
more forms recognised
also loading of unet only sd1.5, sdxl models (for example, saved from Checkpoint Merger)

### DIFF
--- a/backend/loader.py
+++ b/backend/loader.py
@@ -260,7 +260,7 @@ def replace_state_dict(sd, asd, guess):
 
     ##  VAE format 0 (extracted from model, could be sd1, sd2, sdxl, sd3).
     if "first_stage_model.decoder.conv_in.weight" in asd:
-        channels = asd[first_stage_model.decoder.conv_in.weight].shape[1]
+        channels = asd["first_stage_model.decoder.conv_in.weight"].shape[1]
         if model_type == "sd1" or model_type == "sd2" or model_type == "sdxl":
             if channels == 4:
                 for k, v in asd.items():
--- a/backend/loader.py
+++ b/backend/loader.py
@@ -209,12 +209,208 @@ def replace_state_dict(sd, asd, guess):
         for k, v in asd.items():
             sd[vae_key_prefix + k] = v
 
-    if 'text_model.encoder.layers.0.layer_norm1.weight' in asd:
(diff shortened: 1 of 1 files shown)


GlamoramaAttack:
This doesn't work with SD3 medium or SD3.5 large (error messages about no state_dict). I think the changed loader.py file in the backend folderis responsible for it (I see "from backend.diffusion_engine.sd35 import StableDiffusion3" is missing at the start of the script). I was on Version: f2.0.1v1.10.1-previous-652-g184bb04f when it happened. 

Before I updated with git pull to version 652-g184bb04f updates didn't affect SD3/3.5 (it worked at least until updates mid February, after that I didn't use a SD3 or 3.5 model before updating to 652). 

My workaround with a patched loader.py file (following @graemeniedermayer, see below) works, also on version 653-gc4b6fcce as it should with SD3.5. 

---
For anyone who wants to use SD3/3.5 in Forge: additional or patched files can be downloaded from (https://drive.google.com/file/d/1_VYyQ8wQpjh-AoGtWWCa6zK5vEQbwA4K/view?pli=1), those additional files are not part of the packaged Forge versions and some files in the cloned version still aren't updated for SD3/3.5 as the model_list.py one in repositories\huggingface_guess\huggingface_guess which doesn't include class SD3(BASE) or class SD35(BASE). Below is also a link for those who don't want to use other's patched files but rather follow @graemeniedermayer to change them.

-sd35.py in backend\diffusion_engine
-the stable-diffusion-3.5-large script contents from Huggingface/StabilityAI to backend\huggingface\stabilityai\stable-diffusion-3.5-large
-a patched k_prediction.py file to backend\modules
-a patched t5.py file to backend\nn
-a patched classic_engine.py file to backend\text_processing
-a patched loader.py file to backend
-a folder for sd35 Python scripts in modules\models\sd35
-a patched model_list.py file to repositories\huggingface_guess\huggingface_guess

It was discussed on Reddit (url=https://www.reddit.com/r/StableDiffusion/comments/1ge7kpu/sd35_model_on_webui_forge/) following a discussion from @graemeniedermayer in Github (url=https://github.com/lllyasviel/stable-diffusion-webui-forge/pull/2183) about manually changing scripts for it.

PR #2813, opened 2025-04-16
Update postprocessing_upscale.py

wonderinglostsoul44:
Denoise, Sharpen & Histogram Equalization post‑processing.

Tile‑based upscaling helper for huge images without OOM.

Similarity & Suggestion utilities (measure_similarity, suggest_upscale_factor).

Cache key hashing with SHA‑256 and thread safety.

UI controls to select denoise method/strength, sharpen amount, histogram/equalization toggle, and clear cache button.

Improved limit_size_by_one_dimension spelling and logic.

### DIFF
--- a/scripts/postprocessing_upscale.py
+++ b/scripts/postprocessing_upscale.py
@@ -1,6 +1,8 @@
 import re
+import threading
+import hashlib
 
-from PIL import Image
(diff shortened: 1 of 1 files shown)


wonderinglostsoul44:
any reason for the mass rejection rather than intergration, as a community if the OP is inactive we may need to create a new forked project

PR #2877, opened 2025-05-14
add BF16 to GGUF

croquelois:
this PR add the ability to use GGUF model with BF16 layers


croquelois:
@catboxanon would you be able to merge it ?

PR #2915, opened 2025-06-05
Repair ancestrals for flux

croquelois:
repair the ancestrals model for Flux

similar to ComfyUI:
https://github.com/comfyanonymous/ComfyUI/blob/d8759c772bd8ded66e08dac56bb8da636c837c9a/comfy/k_diffusion/sampling.py#L196

@catboxanon are you able to review it ?


maybleMyers:
I can confirm that euler a works now with chroma.

PR #2917, opened 2025-06-06
Fix alpha channel missing

hben35096:
Repair the issue of alpha channel information loss when writing and saving images in webui.

PixPin_2025-06-07_01-46-24


hben35096:
PixPin_2025-07-05_17-58-05

Issue #2945, opened 2025-06-24
Chroma support: IntegratedChromaTransformer2DModel Unexpected: ['scaled_fp8']


hellomrmarky:
I'm using the scaled fp8 versions of Chroma by Clybius
[https://huggingface.co/Clybius/Chroma-fp8-scaled](url)
However these do not seem to work with the lastest integration made.

Console print:
IntegratedChromaTransformer2DModel Unexpected: ['scaled_fp8']

And the output looks like this:

![Image](https://github.com/user-attachments/assets/c2e7c3ae-1516-477d-b24d-ee52c7eadf50)

PR #2946, opened 2025-06-24
add support for fp8 scaled

croquelois:
add support for model with fp8 quantification with scaled weight

https://github.com/lllyasviel/stable-diffusion-webui-forge/issues/2945


iqddd:
How about `.scale_input` and `torch._scale_mm()` for RTX 40/50xx series?

